PR #10, opened 2017-02-13
Adds enum like construct for current RPC status.

MicahZoltu:
This should improve readability by removing the magic numbers, without breaking backwards compatibility.


MicahZoltu:
Investigating build failure.

PR #11, opened 2017-02-14
Adds test backed by TestRPC that shows poor websocket reconnect behavior.

MicahZoltu:
This adds two tests, both backed by ethereumjs-testrpc.  One shows the websocket successfully reconnecting when the server goes away and comes back between calls.  The other shows the behavior (currently bad) when a request is attempted while the server is down.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,4 +6,4 @@ RUN npm install
 
 COPY . /ethrpc
 
-ENTRYPOINT [ "npm", "test" ]
+ENTRYPOINT [ "/ethrpc/node_modules/.bin/mocha" ]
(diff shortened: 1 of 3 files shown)


tinybike:
@MicahZoltu do you still want this PR merged?  It seems like this should be an issue instead, and the PR should be the code that fixes reconnection.

MicahZoltu:
If you have no additional feedback, we can either merge this or close it.  I created this PR primarily to get feedback on the test code I wrote since it is self contained.  If we close the PR, then I'll use this branch for the fix as well.  If we merge it, then I'll create a new branch for the fix.

tinybike:
Let's close the PR then and you can use this branch for the fix.

PR #12, opened 2017-02-18
Refactors transport layer.

MicahZoltu:
This extracts/isolates the transport layer from the rest of the code in ethrpc.  This gives a layer of abstraction that allows each transport type to deal with queueing, reconnecting and errors in isolation without the rest of the code needing to have any understanding about them, their differences or transient connectivity issues.

On connect, one or more will be instantiated and connections will attempted to be established internally.  If at least one transport is working, then the application will proceed.  If no transports are able to be established then connect will fail.

When a request is made, if it fails in a way that is retryable then the transport will retry under-the-hood, reconnecting if necessary, only calling back to the user know once the request has gone through.


index.js
```diff
@@ -130,6 +132,106 @@ module.exports = {
...
+    if (error !== null) return this.configuration.errorHandler(error);
+    if (typeof responseJso !== "object") return this.configuration.errorHandler(new ErrorWithData("Unexpectedly received a message from the transport that was not an object.", responseJso));
+    if (typeof responseJso.id !== "number") return this.configuration.errorHandler(new ErrorWithData("Received a message from the blockchain that didn't have a valid id.", responseJso));
+
+    var outstandingRequest = this.internalState.outstandingRequests[responseJso.id];
+    delete this.internalState.outstandingRequests[responseJso.id];
+    if (typeof outstandingRequest !== "object") return this.configuration.errorHandler(new ErrorWithData("Unable to locate original request for blockchain response.", responseJso));
+    
+    // FIXME: outstandingRequest.callback should be function(Error,object) not function(Error|object)
+    this.parse(responseJso, outstandingRequest.expectedReturnTypes, outstandingRequest.callback);
```
MicahZoltu: @tinybike You indicated that callbacks should all be of the form `function (error, result)` but the callbacks in ethrpc appear to all be `function (errorOrResult)`.  For now I went ahead with keeping the signature the same as I didn't want to do that big of a breaking change, but I did want to call out that fixing it shouldn't be too hard in `ethrpc` with this change.  It will require every call site into `ethrpc` being touched though.  :/

tinybike: That's fine!  Slowly migrating over to the `function (error, result)` syntax, but there's no particular hurry; there's definitely no reason it has to be part of this task.

errors.js
```diff
@@ -0,0 +1,13 @@
+function ErrorWithData(message, data) {
```
tinybike: Digging it

kevinday:
Read through this code this afternoon and this looks likes a really reasonable way to organize this library. Will want @tinybike to give the 👍 before moving forward, but this seems pretty low risk to me.

Was also reading through the rest of ethrpc today and saw there are a bunch of tests for specific augur contract calls. Do we want to port these to optionally use testRPC as well (I don't think this is trivial).

MicahZoltu:
I would assume that anything Augur specific should (eventually) be moved into augur.js?  If that is the case, I have to wonder whether or not it is worth the effort porting those tests over to TestRPC.  At the moment, when I run the tests I do so with `mocha test/testrpc.js test/transporter.js test/transports.js`, explicitly leaving out the RPC tests since they A) fail and B) take forever to run.

Now that I think about it though, the code in this PR will (once applied everywhere) will break all of the RPC tests since none of them call `connect`.  I can either fix the tests by having them call `connect` or rewrite the test to use TestRPC.  The former is quicker as I have not yet figured out how to teach TestRPC about on-chain contracts yet while the latter will give us a more maintainable test suite.

There was also some talk (I think) about making it so the tests could be run either against TestRPC or against a chain, this will take some additional work to setup the infrastructure for, not sure if that should be part of this project or a separate one.

kevinday:
I don't think it makes a ton of sense to have augur specific tests in this repo, but if we do want to keep them I'd be ok with just getting everything back to green however is easiest.

That said I do think it's important to have integration tests *somewhere* that can run optionally against testRPC, so at some point I think it's worth putting in the time to figure out how to bootstrap testRPC with augur contracts/accounts. We can treat that as a separate project, and maybe look to update the tests here if we want to keep them once we have the infrastructure in place for it.

tinybike:
@MicahZoltu This basic organization looks fine to me.  I have a feeling that forgetting to pass that second `null` parameter to `submitRequestToBlockchain` will be a source of errors.  It might be good to set the 2nd parameter to `callback` if it's a function.

Re: Augur-specific tests, those are just in there to exercise the functions associated with contract function calls.  They could definitely be swapped out for contract functions that aren't Augur-specific.  There do need to be tests of the contract-function-call/sendTransaction-related functions, though.

MicahZoltu:
I'm philosophically against allowing users to call functions with incorrect parameters, which is why I set it up to fail fast if they don't provide everything that is expected.  However, it is an easy change to make it support a callback as a second parameter:
```
// allow people to call this function incorrectly rather than failing fast
if (typeof preferredTransport === "function" && !callback) {
  callback = preferredTransport;
  preferredTransport = null;
}
```
so I'll go ahead and make that change.

MicahZoltu:
I'll take some time tomorrow to investigate making contract calls via TestRPC.  Maybe it has a mechanism for letting me stub contract calls with pre-canned responses.  If not, I might be able to monkey-patch TestRPC to allow for this (possibly accompanied by an upstream PR to add the feature to TestRPC).

tinybike:
If TestRPC doesn't support contract calls, we could just stub out the RPC functions for unit tests + have live-chain-only integration tests for actual contract calls.  If TestRPC does support contract calls (or can be made to without a huge amount of effort) that would be ideal, though!

index.js
```diff
@@ -104,51 +77,279 @@ module.exports = {
...
+  excludedFromTxRelay: null,
 
-  txs: {},
+  txs: null,
 
-  rawTxs: {},
+  rawTxs: null,
 
-  rawTxMaxNonce: -1,
+  rawTxMaxNonce: null,
```
tinybike: `rawTxMaxNonce` is set to `null` here, but `resetState` sets it to `-1`.  Is this intentional?

MicahZoltu: Yes.  Using any of the ethrpc functionality without first calling `connect` is an error.  By setting everything to `null` by default, it ensures that the library will fail fast (null reference exception) if the user tries to use it incorrectly.

index.js
```diff
@@ -104,51 +77,279 @@ module.exports = {
...
+   * @param {!object} jso - The JSON-RPC call to make.
+   * @param {?string} transportRequirements - ANY, SYNC or DUPLEX.  Will choose best available transport that meets the requirements.
+   * @param {?function(?Error, ?object):void} callback - Called when a response to the request is received.  May only be null if preferredTransport is SYNC.
+   * @returns {void|?Error|?object} - Returns the error or result if the operation is synchronous.
+   */
+  submitRequestToBlockchain: function (jso, transportRequirements, callback) {
+    var syncErrorOrResult;
+    if (transportRequirements === "SYNC") callback = function (error, result) { return (syncErrorOrResult = (error || result)); };
+
+    // allow people to call this function incorrectly rather than failing fast
```
tinybike: AFAIK this is a well-accepted way to provide a flexible function signature in JS.  The convention that the last argument is the callback (even if you omit some parameters) is used everywhere in the standard library (i.e., Node.js's built-in modules).

MicahZoltu: It is quite common in JavaScript, but it has a minor negative impact on performance due to the way common JavaScript optimizers work (it sees each version of a polymorphic function as a separate thing which can lead to deoptimization) and it results in a signature that is difficult to document/auto-complete.  If you want, I can change the wording of the comment to something else?

index.js
```diff
@@ -104,51 +77,279 @@ module.exports = {
...
+   * @param {!configuration} configuration
+   * @param {!function(?Error):void} initialConnectCallback - if the error parameter is null then the connection was successful
+   * @returns {void}
+   */
+  connect: function (configuration, initialConnectCallback) {
+    this.resetState();
+
+    // overwrite configuration values with user config, throw away unused user config
+    for (var key in this.configuration) {
+      if (configuration[key] === undefined || configuration[key] === null) continue;
```
tinybike: Unless you've omitted it intentionally (why?), there should be an extra check here for inherited properties:
```javascript
if (!this.configuration.hasOwnProperty(key)) continue;
```

index.js
```diff
@@ -104,51 +77,279 @@ module.exports = {
...
+    var syncOnly = !initialConnectCallback;
+    if (syncOnly) initialConnectCallback = function (error) { if (error instanceof Error) throw error; else if (error) throw new ErrorWithData(error); };
+
+    // initialize the transporter, this will be how we send to and receive from the blockchain
+    new Transporter(this.configuration, this.internalState.shimMessageHandler, syncOnly, function (error, transporter) { // jshint ignore:line
+      if (error !== null) return initialConnectCallback(error);
+      this.internalState.transporter = transporter;
+      // ensure we can do basic JSON-RPC over this connection
+      this.version(function (errorOrResult) {
+        if (errorOrResult instanceof Error || errorOrResult.error) return initialConnectCallback(errorOrResult);
```
tinybike: Two questions:
1. Can `errorOrResult` be an actual `Error` object here?
2. Can `errorOrResult` be `null` or `undefined`?  (If so, `errorOrResult.error` will fail.)

index.js
MicahZoltu: It is unnecessary in this case.  `for .. in ..` only iterates over enumerable properties, which means it excludes things like `toString` on the `Object` prototype.  Also, we are iterating over `this.configuration` which we set/control and it is a simple object with no prototype (beyond the default Object prototype).  Since the Object prototype doesn't have any enumerable properties on it, we can be sure that the we will be iterating only over own properties.

index.js
MicahZoltu: Unclear.  I make it a policy of always wrapping errors in `Error` objects, so everywhere in the code I have written, if there is an error it is of type `Error`.  However, I realized (after implementing this originally) that the rest of the code does not follow this pattern everywhere and will instead sometimes return an object that has an `error` property on it.  Rather than trying to figure out all of the places that can return one way or the other, I decided to have all of the checks for errors be like this so they are robust in either case.

It should never be undefined, if it is then there is a programming error.  Null is a more interesting question though, I don't believe it is correct for `version` to return a null result, though I suppose the protocol probably doesn't explicitly state otherwise?  I think the result parser of yours will turn a `null` into an object with an `error` on it though.

If you want, we can add null checking here so this code is resilient to the parser being changed to allow nulls through.  At the moment I don't believe we are at risk of getting a null from this function.

index.js
tinybike: I would like the `hasOwnProperty` check added, but it's not just for consistency.  It's a good idea to have the `hasOwnProperty` check because something else may have added properties to `Object`.  This shouldn't happen in the Augur UI, but it could result in unexpected behavior if ethrpc is used in some other environment where this is true.

tinybike:
I test-drove this a bit tonight and encountered several critical issues:
1. I tried to navigate to the full markets listing (clicked on Markets in the navbar) and the following errors appeared in the browser console:
```
Uncaught TypeError: Cannot read property 'split' of undefined
    at parseCategoricalOutcomes (eval at <anonymous> (http://localhost:8080/main.9ec2eb6ee082f419edb0.js:7671:1), <anonymous>:79:50)
    at eval (eval at <anonymous> (http://localhost:8080/main.9ec2eb6ee082f419edb0.js:7671:1), <anonymous>:52:60)
    at Array.reduce (native)
    at parseOutcomes (eval at <anonymous> (http://localhost:8080/main.9ec2eb6ee082f419edb0.js:7671:1), <anonymous>:39:38)
    at _default (eval at <anonymous> (http://localhost:8080/main.9ec2eb6ee082f419edb0.js:7671:1), <anonymous>:21:41)
    at combination (eval at <anonymous> (http://localhost:8080/assets/scripts/vendor.65a6e0d088c20f34737a.js:2186:1), <anonymous>:119:29)
    at dispatch (eval at <anonymous> (http://localhost:8080/assets/scripts/vendor.65a6e0d088c20f34737a.js:1365:1), <anonymous>:183:22)
    at eval (eval at <anonymous> (http://localhost:8080/main.9ec2eb6ee082f419edb0.js:61:1), <anonymous>:29:7)
    at eval (eval at <anonymous> (http://localhost:8080/assets/scripts/vendor.65a6e0d088c20f34737a.js:1445:1), <anonymous>:14:16)
    at eval (eval at <anonymous> (http://localhost:8080/main.9ec2eb6ee082f419edb0.js:61:1), <anonymous>:20:14)
```
2. Have you been able to successfully use the UI via websockets?  I'm running a local geth instance that has websockets enabled but it shows `augur.subscriptionsSupported=false` in the console.  (I assume this should be `true` if I'm using WS or IPC.)
2a. Chrome's network tab shows a pending websocket connection to 127.0.0.1 that is spamming `eth_getFilterChanges` RPC requests.  Two concerns here: first, I'm not sure this RPC request is even available over websockets.  Second, if using websockets, filters shouldn't be used at all -- all listeners should be subscriptions.
2b. Chrome's network tab also shows HTTP requests sent to https://eth3.augur.net and WS requests sent to wss://ws.augur.net.  My local node is *not* connected to the Ropsten network (I'm on a local, private test chain), so there should be 0 requests directed to eth3.augur.net or ws.augur.net, as these will return conflicting information compared with that of my local node.
3. How do I see which transport I am using with the new ethrpc?  This information needs to be available for debugging.  On the old ethrpc, I could check `wsStatus` and `ipcStatus` to see if I'm connected; what is the equivalent in the new ethrpc?
4. The `broadcast` debugging option needs to behave the same way it did before: it needs to print the RPC requests (exactly as they are sent to geth) to the console, and it needs to print exactly what endpoints the requests are being sent to, and whether the requests are sync or async.
5. When I attempt to navigate to the Trade page (for an individual market), the order book never loads and the order entry inputs never appear.
6. Does this work with MetaMask?  I tried it on both a local test chain and Ropsten, and got the following error:
```
"Error: Unknown address - unable to sign transaction for this address: "0x05ae1d0ca6206c6168b42efcd1fbe0ed144e821b"
    at chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:102634:35
    at chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:102668:5
    at HookedWalletSubprovider.getAccounts (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:3401:11)
    at HookedWalletSubprovider.validateSender (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:102665:8)
    at HookedWalletSubprovider.validateTransaction (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:102632:8)
    at async.waterfall (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:102500:22)
    at nextTask (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:81086:14)
    at Object.waterfall (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:81089:5)
    at HookedWalletSubprovider.handleRequest (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:102499:13)
    at next (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/scripts/background.js:101039:18)"
```
I haven't used MetaMask before -- perhaps it only works on mainnet?

index.js
tinybike: Should probably remove the comment, I think "call this function incorrectly" is a confusing description.

index.js
```diff
@@ -104,51 +77,279 @@ module.exports = {
...
+    this.resetState();
+
+    // overwrite configuration values with user config, throw away unused user config
+    for (var key in this.configuration) {
+      if (configuration[key] === undefined || configuration[key] === null) continue;
+      this.configuration[key] = configuration[key];
+    }
+
+    // validate configuration
+    if (typeof this.configuration.errorHandler !== "function") throw new Error("configuration.errorHandler must be a function");
```
tinybike: Use the `isFunction` function.  (This comment applies in several places below too.)

index.js
```diff
@@ -104,51 +77,279 @@ module.exports = {
...
+   */
+  setupBlockSubscription: function (callback) {
+    var handleSubscribeToNewHeadsResult = function(resultOrError) {
+      if (resultOrError instanceof Error || resultOrError.error) {
+        this.internalState.newBlockIntervalTimeoutId = setInterval(this.getBlockByNumber.bind(this, "latest", false, this.onNewBlock.bind(this, true)), this.BLOCK_POLL_INTERVAL);
+      } else {
+        this.internalState.subscriptions[resultOrError] = this.onNewBlock.bind(this, true);
+      }
+      callback();
+      callback = function () { };
```
tinybike: Why is `callback` reassigned to a noop here?

index.js
```diff
@@ -104,51 +77,279 @@ module.exports = {
...
+      var result = this.subscribeNewHeads(maybeCallback);
+      if (sync) handleSubscribeToNewHeadsResult(result);
+    }.bind(this);
+
+    this.internalState.transporter.addReconnectListener(subscribeToNewBlocks.bind(this, false));
+    clearInterval(this.internalState.newBlockIntervalTimeoutId);
+    subscribeToNewBlocks(!callback);
+  },
+
+  onNewBlock: function (isLatest, block) {
```
tinybike: When would `isLatest` be `false`?

index.js
```diff
@@ -104,51 +77,279 @@ module.exports = {
...
+    }
+    if (this.internalState.lowestSeenSequentialBlock + 1 === newBlockNumber)
+      this.internalState.lowestSeenSequentialBlock = newBlockNumber;
+
+    // FIXME: ethrpc should really store the original block and add getters for making it easier to interact with
+    block.number = newBlockNumber;
+
+    // fetch next missing block if necessary (this will result in recursively catching up)
+    if (this.internalState.lowestSeenSequentialBlock != this.internalState.lastSeenLatestBlockNumber)
+      this.getBlockByNumber(this.internalState.lowestSeenSequentialBlock + 1, false, this.onNewBlock.bind(this, false));
```
tinybike: What is the purpose of getting all the blocks?  (Is this intended to deal with node disconnects?)  It seems like it would be simpler to just call `eth_blockNumber` then `eth_getBlockByNumber` once.

index.js
MicahZoltu: Fixed.

index.js
MicahZoltu: Fixed.

index.js
MicahZoltu: Fixed.  Why though?  `typeof variable === "function"` is significantly faster than checking the string result of the prototype.

index.js
MicahZoltu: I added a comment (should have done this originally) to make it clear why this is happening.  The issue is that the same logic is used for processing both initial connect (which is when the callback makes sense) and reconnects (when calling the callback doesn't make sense).  By clearing the callback after calling it the first time (on initial connect) we can otherwise re-use all of the code for block subscription management on reconnect.

index.js
MicahZoltu: If you are fetching a missed block, you will call this with `isLatest = false` but in such a scenario you don't want to update the the current state (this.block) to the received block.  Your next comment is relevant to this, lets discuss the merits of this there.

index.js
MicahZoltu: I originally wrote the code for fetching missed blocks before we had all of the conversations around whether or not we should be backfilling subscriptions (blocks included).  I believe at the moment, only the latest block is actually used.  I decided to leave the code in because I thought it would be valuable later if/when we decided we did want to ensure all blocks were seen.  However, if you think we'll be taking a different route for missed blocks (or perhaps we just don't care about missed blocks) then I suspect a lot of this code could be deleted and replaced with something much simpler like renaming `onNewBlock` to `onNewLatestBlock` and not calling it for anything other than latest blocks.

Let me know how you want to proceed with regards to this code and I'll take the appropriate action.

index.js
tinybike: Good question.  It was because this used to be true:
```javascript
typeof /myregex/ === "function"
```
But looking this up it seems like this is no longer true in modern browsers.  So actually `typeof variable === "function"` should be fine now!

index.js
tinybike: > However, if you think we'll be taking a different route for missed blocks (or perhaps we just don't care about missed blocks) then I suspect a lot of this code could be deleted and replaced with something much simpler like renaming onNewBlock to onNewLatestBlock and not calling it for anything other than latest blocks.

Let's do this for now.  If we end up wanting to use this for backfills, this code is now saved via git so we can copy-and-paste it easily enough if needed.

index.js
tinybike: I prefer to avoid functions that behave differently depending on the number of times they're called.  Whether the initial connect has happened yet could be inferred from whether `rawTxMaxNonce` (and other props) are non-null, or, alternatively, it might be worth creating a boolean `isConnected` prop for readability.

MicahZoltu:
> I haven't used MetaMask before -- perhaps it only works on mainnet?

When testing MetaMask, did you move the augur connect code into `onload`?

MetaMask support is tenuously tested at best.  I added support for it and did some *basic* testing , but I wouldn't be surprised if it was buggy.  At the moment, my primary concern is getting my outstanding branches merged because they are already quite large and I really am not comfortable with large long running pull requests.  Since the MetaMask support is new and doesn't impact existing code I put off doing rigorous testing for the time being.  If you like, we can wait to merge this until after MetaMask is fully verified but I'm worried that may take a while since I'm not particularly familiar with it and debugging it is likely going to be a bit of a pain.  The only reason I added the MetaMask stuff into this branch/PR was because the change was so small, it didn't impact the rest of the code much at all and I wanted to verify it would fit in the current transport model.  If you want, I can extract the MetaMask code into a separate PR?

> Have you been able to successfully use the UI via websockets? I'm running a local geth instance that has websockets enabled but it shows augur.subscriptionsSupported=false in the console. (I assume this should be true if I'm using WS or IPC.)
> * Chrome's network tab shows a pending websocket connection to 127.0.0.1 that is spamming eth_getFilterChanges RPC requests. Two concerns here: first, I'm not sure this RPC request is even available over websockets. Second, if using websockets, filters shouldn't be used at all -- all listeners should be subscriptions.
> * Chrome's network tab also shows HTTP requests sent to https://eth3.augur.net and WS requests sent to wss://ws.augur.net. My local node is not connected to the Ropsten network (I'm on a local, private test chain), so there should be 0 requests directed to eth3.augur.net or ws.augur.net, as these will return conflicting information compared with that of my local node.

I just fixed a bug in the subscription detection code in augur.js, I think I introduced it in one of my last changes and didn't go back and retest web sockets, that was my bad.

All requests that work with HTTP also work over WS, though you are correct that it shouldn't have been taking the `getFilterChanges` code path when subscriptions were supported.  That is the bug I fixed.

I'm not seeing any contact with `eth3.augur.net` or `ws.augur.net` when I connect to a local node.  At the moment, the way things are hooked up, if you give it no websocket node to talk to but you tell it to use fallbacks then it will fallback to `ws.augur.net`.  The same is true for `eth3.augur.net`.  Alternatively, if it is unable to connect to the URL you provided then it will fallback to hosted nodes for either of those.  It sounds like something is causing you to hit the fallback logic?  What `env.json` are you using?  Note, at the moment the the fallback nodes are hard coded to `eth3.augur.net` and `ws.augur.net` here https://github.com/ethereumjs/ethereumjs-connect/blob/ethrpc-update/index.js#L229-L230 which seems like the wrong place to add fallbacks, but it required the least changes.

> How do I see which transport I am using with the new ethrpc? This information needs to be available for debugging. On the old ethrpc, I could check wsStatus and ipcStatus to see if I'm connected; what is the equivalent in the new ethrpc?

`augurjs.augur.rpc.internalState.transporter.internalState` contains the transports that are currently being used.  You can look at `.metaMaskTransport`, `.httpTransport`, `.wsTransport` and `.syncTransport` on that object to find out which are available.  If it is null, it means that transport isn't available in this session.  Note: Those things shouldn't be used outside of debugging, which is why they are in the `internalState` object.

> The broadcast debugging option needs to behave the same way it did before: it needs to print the RPC requests (exactly as they are sent to geth) to the console, and it needs to print exactly what endpoints the requests are being sent to, and whether the requests are sync or async.

I have added this in, so if `.debug.broadcast === true` then it will print out the request and the destination.
Professional opinion: there shouldn't be debugging code in production (master) and instead a debugger should be used if you need to inspect this data while debugging.  In this case in particular, all of the information you want is available in Chrome's networking tab.

> Running the UI's test suite (yarn test) shows that all gas fees in transactions are now set to 0, instead of their expected values.

This is because the tests are using ethrpc without first calling connect (as mentioned elsewhere in this PR, this is incorrect usage of the library).  The quick-fix for this would be to have the tests set `rpc.gasPrice = 20000000000`.  This would make the tests pass, though it is unclear what exactly the tests are testing if this solution is employed.  Let me know if you want me to get a PR out that applies this, or if you want to explore other options for addressing this issue.

index.js
MicahZoltu: Fixed.  I restructured the function a bit, recommend re-reviewing `setupBlockSubscription(...)`.

index.js
MicahZoltu: Fixed.

tinybike:
> This is because the tests are using ethrpc without first calling connect (as mentioned elsewhere in this PR, this is incorrect usage of the library). The quick-fix for this would be to have the tests set rpc.gasPrice = 20000000000. This would make the tests pass, though it is unclear what exactly the tests are testing if this solution is employed. Let me know if you want me to get a PR out that applies this, or if you want to explore other options for addressing this issue.

These tests are of the trading preview functions.  They use the gasprice to estimate the amount of ether that will be needed to run the transaction, prior to actually sending it.  It's good to have a nonzero gasprice value since the tests are checking that the numbers come out as expected, so I'll just set it for the tests.

tinybike:
Okay, so the rpc.gasPrice value was straightforward to mock in augur.js's tests.  However, I'm for some reason not able to mock this value using the elaborate mocking system (w/ proxyquire) in the UI's tests.  @JohnDanz, do you have any idea how to properly mock augur.js which imports ethereumjs-connect which imports ethrpc, which has a property called `gasPrice` that needs to be mocked to `20000000000`?  I tried "nested proxyquires" but for some reason the `20000000000` value gets set briefly, then immediately overwritten with a `null`.

tinybike:
More UI test errors being thrown:
```

  1) modules/create-market/selectors/create-market-form.js step 5 should deliver the correct values to components:
     BigNumber Error: times() not a number: null
      at raise (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1181:25)
      at /home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1169:33
      at new BigNumber (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:195:67)
      at BigNumber.P.times.P.mul (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:2083:37)
      at Augur.calculateValidityBond (/home/jack/src/augur.js/src/modules/abacus.js:95:69)
      at Object.select (src/modules/create-market/selectors/form-steps/step-5.js:11:30)
      at Context.<anonymous> (test/create-market/selectors/create-market-form-test.js:261:24)

  2) modules/create-market/selectors/form-steps/step-5.js select should call the correct action onSubmit:
     BigNumber Error: times() not a number: null
      at raise (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1181:25)
      at /home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1169:33
      at new BigNumber (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:195:67)
      at BigNumber.P.times.P.mul (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:2083:37)
      at Augur.calculateValidityBond (/home/jack/src/augur.js/src/modules/abacus.js:95:69)
      at Object.select (src/modules/create-market/selectors/form-steps/step-5.js:11:30)
      at Context.<anonymous> (test/create-market/selectors/form-steps/step-5-test.js:50:31)

  3) modules/create-market/selectors/form-steps/step-5.js select should return the correct object for a binary market WITH an order book:
     BigNumber Error: times() not a number: null
      at raise (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1181:25)
      at /home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1169:33
      at new BigNumber (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:195:67)
      at BigNumber.P.times.P.mul (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:2083:37)
      at Augur.calculateValidityBond (/home/jack/src/augur.js/src/modules/abacus.js:95:69)
      at Object.select (src/modules/create-market/selectors/form-steps/step-5.js:11:30)
      at Context.<anonymous> (test/create-market/selectors/form-steps/step-5-test.js:94:31)

  4) modules/create-market/selectors/form-steps/step-5.js select should return the correct object for a binary market WITHOUT an order book:
     BigNumber Error: times() not a number: null
      at raise (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1181:25)
      at /home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1169:33
      at new BigNumber (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:195:67)
      at BigNumber.P.times.P.mul (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:2083:37)
      at Augur.calculateValidityBond (/home/jack/src/augur.js/src/modules/abacus.js:95:69)
      at Object.select (src/modules/create-market/selectors/form-steps/step-5.js:11:30)
      at Context.<anonymous> (test/create-market/selectors/form-steps/step-5-test.js:284:31)

  5) modules/create-market/selectors/form-steps/step-5.js select should return the correct object for a categorical market WITH an order book:
     BigNumber Error: times() not a number: null
      at raise (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1181:25)
      at /home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1169:33
      at new BigNumber (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:195:67)
      at BigNumber.P.times.P.mul (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:2083:37)
      at Augur.calculateValidityBond (/home/jack/src/augur.js/src/modules/abacus.js:95:69)
      at Object.select (src/modules/create-market/selectors/form-steps/step-5.js:11:30)
      at Context.<anonymous> (test/create-market/selectors/form-steps/step-5-test.js:426:31)

  6) modules/create-market/selectors/form-steps/step-5.js select should return the correct object for a categorical market WITHOUT an order book:
     BigNumber Error: times() not a number: null
      at raise (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1181:25)
      at /home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1169:33
      at new BigNumber (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:195:67)
      at BigNumber.P.times.P.mul (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:2083:37)
      at Augur.calculateValidityBond (/home/jack/src/augur.js/src/modules/abacus.js:95:69)
      at Object.select (src/modules/create-market/selectors/form-steps/step-5.js:11:30)
      at Context.<anonymous> (test/create-market/selectors/form-steps/step-5-test.js:649:31)

  7) modules/create-market/selectors/form-steps/step-5.js select should return the correct object for a scalar market WITH an order book:
     BigNumber Error: times() not a number: null
      at raise (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1181:25)
      at /home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1169:33
      at new BigNumber (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:195:67)
      at BigNumber.P.times.P.mul (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:2083:37)
      at Augur.calculateValidityBond (/home/jack/src/augur.js/src/modules/abacus.js:95:69)
      at Object.select (src/modules/create-market/selectors/form-steps/step-5.js:11:30)
      at Context.<anonymous> (test/create-market/selectors/form-steps/step-5-test.js:792:31)

  8) modules/create-market/selectors/form-steps/step-5.js select should return the correct object for a scalar market WITHOUT an order book:
     BigNumber Error: times() not a number: null
      at raise (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1181:25)
      at /home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:1169:33
      at new BigNumber (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:195:67)
      at BigNumber.P.times.P.mul (/home/jack/src/augur.js/node_modules/bignumber.js/bignumber.js:2083:37)
      at Augur.calculateValidityBond (/home/jack/src/augur.js/src/modules/abacus.js:95:69)
      at Object.select (src/modules/create-market/selectors/form-steps/step-5.js:11:30)
      at Context.<anonymous> (test/create-market/selectors/form-steps/step-5-test.js:987:31)
```

MicahZoltu:
When I looked at those earlier, I believe they were all the same issue (gasPrice not mocked).  Based on other comments and commit, looks like the path you want to take is setting gasPrice in the tests, so I'll take care of that tomorrow for all of these tests (or otherwise fix them if I am wrong about the gasPrice being the problem).

tinybike:
> When I looked at those earlier, I believe they were all the same issue (gasPrice not mocked). Based on other comments and commit, looks like the path you want to take is setting gasPrice in the tests, so I'll take care of that tomorrow for all of these tests (or otherwise fix them if I am wrong about the gasPrice being the problem).

Yep, adding a fallback value to relevant files in the UI and augur.js fixed it.

tinybike:
Btw, if you're able to get the mocks working, that's preferable for the UI `rpc.gasPrice` references -- `augur.getTradingActions` should work without ethrpc being connected, but UI-without-ethrpc is probably not going to be a thing.

tinybike:
Alright, websockets are working for me!  I'm still seeing a few HTTP and websocket requests directed at https://eth3.augur.net and wss://ws.augur.net, while connected to a local node (on network 10101, my local test network).  My env.json has fallback set to true so that's probably the issue:
```json
{
  "gethHttpURL": "http://127.0.0.1:8545",
  "gethWebsocketsURL": "ws://127.0.0.1:8546",
  "hostedNodeFallback": true,
  "augurNodeURL": null,
  "autoLogin": false,
  "branchID": null,
  "reportingTest": false,
  "debug": {
    "trading": true,
    "reporting": true,
    "broadcast": false
  },
  "fundNewAccountFromAddress": {
    "address": "0x05ae1d0ca6206c6168b42efcd1fbe0ed144e821b",
    "amount": 100
  }
}
```
Setting `hostedNodeFallback=false` causes the stray requests to the network 3 nodes to stop, hurrah.  A couple of thoughts:
1. We probably should check to make sure the network ID hasn't changed when the user hops between nodes.  If the user changes networks (blockchains) then almost all the data in ethrpc and augur.js will need to be reset (the transactions etc.).  It's probably easiest/safest to just not allow this to happen at all.
2. I agree that ethereumjs-connect is not the right place for the hosted node URLs to be stored.  I think a better place for them would be the UI's env.json file(s).  `gethHttpURL` and `gethWebsocketsURL` could optionally be arrays -- the first element would be the first endpoint checked, the others would be fallbacks and could include eth3.augur.net and ws.augur.net in the default env.json.

tinybike:
> When testing MetaMask, did you move the augur connect code into onload?

Ah, no I didn't -- I didn't realize I needed to.  I'll test it with the connect code moved tomorrow.  It's not essential for MetaMask to be part of this PR, but it's a nice bonus if it works.

tinybike:
> Professional opinion: there shouldn't be debugging code in production (master) and instead a debugger should be used if you need to inspect this data while debugging. In this case in particular, all of the information you want is available in Chrome's networking tab.

Point taken.  I do usually use the networking tab myself.  I like having the debug logs just printing to the console because I can ask anyone using the site to just hit CTRL+SHIFT+J and tell me if there's a bunch of red stuff in the console (or scroll up until you see blue text, etc.).

MicahZoltu:
> We probably should check to make sure the network ID hasn't changed when the user hops between nodes. If the user changes networks (blockchains) then almost all the data in ethrpc and augur.js will need to be reset (the transactions etc.). It's probably easiest/safest to just not allow this to happen at all.

Very good point, I didn't consider that the user may have supplied URLs for different transports that may go to different networks.  The `Transporter` would be the right place to validate that the network is the same for all connected transports.  What should the logic be when there is a conflict?  Fail connect entirely?  Pick the network with the most connections (assuming there is one)?  Pick a random network?  Pick a network based on preferred transport order (MetaMask > IPC > WS > HTTP)?

I propose that we don't actually do that in this PR and instead leave it for a separate PR.  I think there is value in getting this PR tied off nicely and I don't believe this is a regression (the old code, I believe, would let you connect to different networks via WS/IPC/HTTP).

MicahZoltu:
> I'm still seeing a few HTTP and websocket requests directed at https://eth3.augur.net and wss://ws.augur.net, while connected to a local node (on network 10101, my local test network).

Oh!  I know what these are.  During initial connect, ethrpc will do a test connection to all supplied nodes to find out which ones are available for connecting to.  Since you have fallback enabled, this means it will test the connectivity of the fallback nodes.  However, assuming that you get a connection to `localhost`, the fallback node connections will be dropped after the initial connect since a "better" node was found.

MicahZoltu:
> I agree that ethereumjs-connect is not the right place for the hosted node URLs to be stored. I think a better place for them would be the UI's env.json file(s). gethHttpURL and gethWebsocketsURL could optionally be arrays -- the first element would be the first endpoint checked, the others would be fallbacks and could include eth3.augur.net and ws.augur.net in the default env.json.

Fixed.  Augur UI now deals with the fallback logic directly.  PR: https://github.com/AugurProject/augur/pull/878

tinybike:
> I propose that we don't actually do that in this PR and instead leave it for a separate PR. I think there is value in getting this PR tied off nicely and I don't believe this is a regression (the old code, I believe, would let you connect to different networks via WS/IPC/HTTP).

Agreed.  I think the way to handle the network ID check is to assume that the first network connected to is the one that the user intended to connect to.  If ethrpc subsequently jumps to a different node with a different network ID, it should probably just ignore that node and try the next one (or fail if there are no remaining nodes to try).

tinybike:
I'm running a final set of tests, the "reporting sequence" -- spawn a new branch, make 3 markets, create order books, trade in the markets, wait for expiration, report on the markets, close the markets, verify the outcomes are set, verify rep updates correctly, verify payouts are correct.  I noticed that `state.blockchain.currentBlockNumber` is not updating as expected in the UI.  I think the block (`newHeads`) subscription set up by augur.js's filters.js isn't working / new block notifications aren't propagating correctly to the UI:
```js
> state.blockchain.currentBlockNumber
1716
> augurjs.augur.rpc.block.number
1732
```
Those numbers should be the same, since the UI has an action that autotriggers any time a new block arrives that should keep the number updated.

tinybike:
I just hopped back to the master branches of augur/augur.js/ethereumjs-connect/ethrpc and verified that this was working before.  Looking over the filters.js code, one non-obvious thing that jumps out at me is that the augur.js filters `newHeads` subscription is a *different* `newHeads` subscription than the one used internally by ethrpc -- possibly this causes an issue in the new ethrpc?  (It would actually be better (more efficient) to make them use the same `newHeads` subscription.)

MicahZoltu:
I'll look into it.  The block subscription mechanism in ethrpc is (now) completely isolated from any other subscriptions that augur.js may setup.  Assuming Augur.js is setting up and managing its own subscriptions, then the ethrpc changes _shouldn't_ have changed any functionality.  If augur.js was poking into the internals of ethrpc to try and leverage its internal subscriptions though, then it may have broken if I moved something that it was depending on.  I'll let you know what I find either way.

MicahZoltu:
I found/fixed the bug (it was in ethrpc subscription registration).  Though, I'm a tad surprised to find that augurjs filters are always about a block behind ethrpc's block.  Debugging a little bit, I found that both were receiving subscription updates at the same time with the same block, but `state.blockchain.currentBlockNumber` was always the number of the *previous* block, not the current block.

Debugging a bit more, it appears that the augur subscription is just a trigger that causes `syncBlockchain` to be fired.  This function goes and just copies the block number found in `rpc.block.number` into augurjs.  However, since both are receiving subscription updates at the same time, it is undefined whether ethrpc will have updated its block number to latest before augurjs handles the subscription update and copies the value from ethrpc.  This results in augurjs often copying a block number over that is stale (previous block) whereas if it waited for just a moment, in most cases it would get the current block number.

I'm reasonably confident this was an existing bug so my plan is to not address it in this PR, but I wanted to make sure you were aware of it.  Other than this, block subscriptions should be working correctly now.

tinybike:
Oh interesting, yes you're right re: the block.number -> blockchain.currentBlockNumber race condition.  I'll fix that separate from this PR.  Nice catch!

tinybike:
Okay, finished the full reporting sequence successfully!  Everything looks good; merging.

PR #14, opened 2017-03-07
Raw transactions

tinybike:


tinybike:
Requesting a review from @MicahZoltu -- you'll need to glance over this anyway when we're merging your PR so might as well review it :grin:

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+    packaged.value = payload.value || "0x0";
+    if (payload.gasLimit) {
+      packaged.gasLimit = abi.hex(payload.gasLimit);
+    } else if (this.block && this.block.gasLimit) {
+      packaged.gasLimit = abi.hex(this.block.gasLimit);
+    } else {
+      packaged.gasLimit = this.DEFAULT_GAS;
+    }
+    if (this.networkID && !isNaN(parseInt(this.networkID, 10)) && parseInt(this.networkID, 10) < 109) {
+      packaged.chainId = parseInt(this.networkID, 10);
```
MicahZoltu: Why do we care about chain ID in this code?

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+    packaged.nonce = payload.nonce || "0x0";
+    packaged.value = payload.value || "0x0";
+    if (payload.gasLimit) {
+      packaged.gasLimit = abi.hex(payload.gasLimit);
+    } else if (this.block && this.block.gasLimit) {
+      packaged.gasLimit = abi.hex(this.block.gasLimit);
+    } else {
+      packaged.gasLimit = this.DEFAULT_GAS;
+    }
+    if (this.networkID && !isNaN(parseInt(this.networkID, 10)) && parseInt(this.networkID, 10) < 109) {
```
MicahZoltu: What is the magic number `109` for?  Where does it come from?  Why are we comparing against it?

MicahZoltu:
Travis appears to be choking on one of the tests, you may want to investigate that.

tinybike:
eth_signTransaction isn't a thing as far as I know.  We have to sign in the
browser because geth does not have the signing key.  Re: nonce.  It is just
the number of transactions completed by this account.  However due to
pending or simultaneous transactions it can be wrong, in which case it
needs to simply be incremented and retried.  chainId is part of the updated
RPC spec.  The magic number 109 is because chainId higher than this will
cause the v param (in ECDSA sig) to exceed 1 byte and fail.  Validation can
fail due to insufficient gas.  Ethereumjs-tx is ES6 so we babelify it,
exactly as other ES6 libs are handled.

tinybike:
Ah, I forgot to update Travis yml file so it's using the old tests, will update that.

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+        delete packaged.nonce;
+        return this.setNonceThenSubmitRawTransaction(packaged, address, privateKey, callback);
+      }
+      if (!isFunction(callback)) return rawTransactionResponse;
+      return callback(rawTransactionResponse);
+    }
+
+    // rawTransactionResponse is the txhash if nothing failed immediately
+    // (even if the tx is nulled, still index the hash)
+    this.rawTxs[rawTransactionResponse] = {tx: packaged, cost: abi.unfix(cost, "string")};
```
tinybike: `abi` is generic Ethereum stuff.

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+        var err = clone(errors.RLP_ENCODING_ERROR);
+        err.bubble = rawTransactionResponse;
+        err.packaged = packaged;
+        if (!isFunction(callback)) return err;
+        return callback(err);
+      } else if (rawTransactionResponse.message.indexOf("Nonce too low") > -1) {
+        if (this.debug.broadcast || this.debug.nonce) {
+          console.info("[ethrpc] Nonce too low, incrementing:", rawTransactionResponse.message, packaged, this.rawTxMaxNonce);
+        }
+        ++this.rawTxMaxNonce;
```
tinybike: You can lookup number of transactions, which isn't really more likely to be correct than simple increment.  So better to increment and save an RPC request.

index.js
tinybike: It's being unfixed (converted to base 10).  It's previously in fixed point.  The string part is incidental.  Please lookup how abi/unfix/etc work before making recommendations about them!

MicahZoltu:
Ah, the key is in the browser is the part that slipped my mind.  There is a [`eth_signTransaction`](https://github.com/ethcore/parity/wiki/JSONRPC-eth-module#eth_signtransaction), but it signs it with keys controlled by Geth, not by the browser.  The reasoning for this code makes more sense now.

index.js
MicahZoltu: Is the name `augur-abi` just a leftover from a time before it was generalized?

MicahZoltu: > It's being unfixed (converted to base 10). It's previously in fixed point.

This doesn't align with the code.  https://github.com/AugurProject/augur-abi/blob/master/test/abi.js#L956-L959
That test suggests that given a hex value, it will assume that it is a floating point value and convert it to such and store that result in a string.  Why is cost a floating point number to begin with, it feels like it should be a fixed point number.  Why are we converting it to a string representation instead of leaving it as a hex encoded octuple-precision floating point value?

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+    // nonce ok, complete sequence
+    if (!isFunction(callback)) return rawTransactionResponse;
+    callback(rawTransactionResponse);
+  },
+
+  submitRawTransaction: function (packaged, address, privateKey, callback) {
+    for (var rawTxHash in this.rawTxs) {
+      if (!this.rawTxs.hasOwnProperty(rawTxHash)) continue;
+      if (this.rawTxs[rawTxHash].tx.nonce === packaged.nonce && (!this.txs[rawTxHash] || this.txs[rawTxHash].status !== "failed")) {
+        packaged.nonce = abi.hex(this.rawTxMaxNonce + 1);
```
tinybike: Micah the way the nonce stuff works is that it keeps track of the highest nonce client-side so that if you dump a bunch of transactions thru ethrpc at the same time they won't all be assigned identical nonces and have to be retried.

Having nonce holes does cause a problem.  If you skip a nonce, geth will just sit there stupidly waiting for the "missing" nonce forever, and won't return any kind of error or other indication that you've done something wrong.  So the application will literally just silently stop functioning if you skip a nonce.  This is why the nonce stuff is over-engineered to the nth degree to make sure that if we make mistakes with nonces, they're too *low* rather than too high.  Low nonces are fine -- they trigger either a "nonce too low" error, a "duplicate transaction" error, or the transaction gets nulled out.  Either way it's something you can notice and react to.

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+      return callback(errors.RAW_TRANSACTION_ERROR);
+    }
+    if (rawTransactionResponse.error) {
+      if (rawTransactionResponse.message.indexOf("rlp") > -1) {
+        var err = clone(errors.RLP_ENCODING_ERROR);
+        err.bubble = rawTransactionResponse;
+        err.packaged = packaged;
+        if (!isFunction(callback)) return err;
+        return callback(err);
+      } else if (rawTransactionResponse.message.indexOf("Nonce too low") > -1) {
```
tinybike: I'd like a better way of doing this, but AFAIK this message is the only indication from geth as to what specific problem happened.  The associated error code (-32000) can be triggered by many different things, not just nonce too low.

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+
+  parseRawTransactionResponse: function (rawTransactionResponse, packaged, address, privateKey, cost, callback) {
+    if (this.debug.broadcast) {
+      console.debug("[ethrpc] sendRawTransaction response:", rawTransactionResponse);
+    }
+    if (!rawTransactionResponse) {
+      if (!isFunction(callback)) return errors.RAW_TRANSACTION_ERROR;
+      return callback(errors.RAW_TRANSACTION_ERROR);
+    }
+    if (rawTransactionResponse.error) {
```
tinybike: Generally I try to avoid nested functions in code that sees a lot of use for performance reasons (creating functions is expensive in JS).  Some of the other ugliness you commented on is like that for the same reason -- e.g., raw for loops instead of `forEach` -- just because it's faster and this code gets called a lot.

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+    this.rawTxs[rawTransactionResponse] = {tx: packaged, cost: abi.unfix(cost, "string")};
+
+    // nonce ok, complete sequence
+    if (!isFunction(callback)) return rawTransactionResponse;
+    callback(rawTransactionResponse);
+  },
+
+  submitRawTransaction: function (packaged, address, privateKey, callback) {
+    for (var rawTxHash in this.rawTxs) {
+      if (!this.rawTxs.hasOwnProperty(rawTxHash)) continue;
```
tinybike: `rawTxs` is actually an object here, not an array.

index.js
tinybike: @MicahZoltu first of all, fix/unfix are used all throughout the codebase.  If they were incorrect, all values from the application would be wrong.  So you should probably start from the assumption that this stuff all works and you are looking for the gap in your understanding.

That hex value is a hex-encoded fixed-point number.  Like every other value we get back from the contracts it is an int256.  To show to users / to work with conveniently in JavaScript we would like it to be a normal base10 number, i.e., not fixed-point.  However we do not want to use JS floating point because binary floating point results in rounding errors.  Instead we deal with the numbers as stringified base10 numbers.

Demonstration that what I'm saying is correct:
```
> abi.fix("3.1415", "hex")
'0x2b98d99b09e3c000'
```
This is the same result we get from:
```
> new BigNumber("3.1415", 10).times(new BigNumber(10, 10).toPower(18)).toString(16)
'2b98d99b09e3c000'
```
And if we use `abi.unfix` we should get back the (stringified) float that we started with, `3.1415`:
```
> abi.unfix("0x2b98d99b09e3c000", "string")
'3.1415'
```

index.js
```diff
@@ -1448,14 +1447,179 @@ module.exports = {
...
+      if (this.rawTxs[rawTxHash].tx.nonce === packaged.nonce && (!this.txs[rawTxHash] || this.txs[rawTxHash].status !== "failed")) {
+        packaged.nonce = abi.hex(this.rawTxMaxNonce + 1);
+        if (this.debug.broadcast || this.debug.nonce) {
+          console.debug("[ethrpc] duplicate nonce, incremented:", parseInt(packaged.nonce, 16), this.rawTxMaxNonce);
+        }
+        break;
+      }
+    }
+    if (parseInt(packaged.nonce, 16) <= this.rawTxMaxNonce) {
```
tinybike: It can't be passed to sendRawTransaction with a number.  All that gets passed to sendRawTransaction is the signed hex data.  The nonce (and everything else) has to be correctly formatted prior to being signed.  I agree there's too many conversions though.

PR #17, opened 2017-03-15
Creates source directory and moves files into it.

MicahZoltu:
This PR is in preparation for some additional work that I will be doing on ethrpc.  Its primary purpose is to move all source code into a `source` directory, rather than having it spread out across root and subdirectories.

In the process of doing this, I found that the linter wasn't actually linting the entire project.  Upon fixing that, I find several linting errors (some of which caught real bugs) which are also fixed in this PR.

I disabled `laxbreak` linter error partly because it is deprecated: http://jshint.com/docs/options/#laxbreak and partly because of the reasons outlined in [this comment](http://stackoverflow.com/questions/15140740/explanation-of-jshints-bad-line-breaking-before-error#comment39683849_15141168) and [this comment](http://stackoverflow.com/questions/15140740/explanation-of-jshints-bad-line-breaking-before-error#comment41169545_15140740).  Research suggests the reason it originally made it into JSHint (which tries to be un-opinionated) is because there was FUD around the automatic semicolon inserter, but it turns out this linting option doesn't address that at all, which is why I believe it is being deprecated.

I have verified that all of the gulp and NPM tasks still work correctly (which includes all tests passing).


tinybike:
Couple of minor thoughts:
1. Suggest using `src` instead of `source` (to match our other repos).
2. Good catch re: linter not linting everything.  I recently implemented a more aggressive linting scheme in augur.js.  It is not yet applied to the other middleware repos but I think the middleware would benefit from more thorough linting in general.  Thoughts?  It is basically Crockford's JSHint stuff plus a few stylistic things: https://github.com/AugurProject/augur.js/blob/master/.eslintrc

Related to (2): I'd like to adapt the MISRA-C guidelines for JavaScript, since these have been used with good effect to write reliable safety-critical software for many years.  Obviously not all the rules are applicable!  Here is the subset that seems relevant to me: http://socrates.io/#ap6nmdc

MicahZoltu:
Changed `source` to `src`.

Regarding https://github.com/AugurProject/augur.js/blob/master/.eslintrc, switching to that is out of scope of this PR I think, but in general I'm a fan of consistency (even if I don't agree with the pattern being followed).  Looking at http://socrates.io/#ap6nmdc, most of it seems reasonable and good practice.  A few of them will result in a bit more code, though in most cases I think it is worth the cost (e.g., loop variables not being modified in loop).  However, I have a history of having a pretty strong contention with most of section 14 (Control Flow).  I have heard the arguments for many of these (like single exit point from functions) and the net effect is increased nesting, increased cognitive load on reader, decreased readability, and in general code that is incredibly difficult to read/maintain.  Also, 16.2 (no recursion) is a bit overly strict.  While in general I do advocate for avoiding recursion when possible, sometimes it really is the "best" solution to a problem.

As for ESLint, there are a handful that I disagree with, though in most cases (as I mentioned above) consistency within a code-base is more important than preference.  For example, I think that since JavaScript doesn't have a concept of private methods, allowing for nested function declarations can provide a valuable logical grouping of functionality and encapsulation of sub-logic without exposing implementation details to the user.  Also, since nested functions can have captures there can be some big benefits on code readability and reduction by having nested functions.  However, I wouldn't throw a huge fit over that being added to a linter.  😄

Finally, https://github.com/AugurProject/augur.js/blob/master/.eslintrc: as previously consistency is valuable and therefore most of my stylistic disagreements are rather meaningless.  There are a couple however that you have added beyond the recommended set that I think *hurt* rather than help.  `vars-on-top` doesn't add any value in the face of other linter checks like `no-shadow` and it moves variable declaration *away* from where it is used.  By separating declaration and initialization, you not only increase the cognitive load burden on the user, but you also increase the liklihood that the variable will not be assigned.  The linter can often check this for you, but since it doesn't add any value and it has other negative side effects, I think it is a bad idea.  In general, code that has "vars at top" ends up being far less readable IMO.  In most languages that support non-vars-at-top (i.e., everything but `C`) the recommendation is to declare variables when you initialize them, never before (unless you absolutely have to) as this prevents bugs caused by forgetting to initialize a variable.

I think realistically, if you are interested in writing robust JavaScript and you are willing to sink the time into converting your projects over to it your best bet is to write in a static language and compile to JavaScript.  TypeScript, IMO, is the leading contender here as it supports full static typing, works on all platforms, has an abundance of tooling for it, is a superset of ECMAScript, and (potentially most importantly) has a concept of non-null types.  I have heard some people argue that the transpilation step makes the JavaScript harder to read, but I argue that if you are reading the output JavaScript then you are doing it wrong.  I know I sure as hell never (well, almost never) read the assembly output of my C/C++/Go/Rust code, and I never (almost) read the CIL output of my C# or my Java Bytecode output of my Java/Kotlin.  Let the compiler do its job (it does it quite well) and only look at the original source.  Treat the output JS as binary files.  Tooling in the JavaScript ecosystem has advanced to the point where basically everything supports Source Maps and you should *always* have source maps available.

MicahZoltu:
Hmm, build failures seem to be related to Istanbul and it claiming to not be able to find this repo.  Does something need to happen with Istanbul config to hook it up to this repo now that it has moved?  I updated all references to `AugurProject` in the repo to point at the new location, but that doesn't appear to have helped.

PR #18, opened 2017-03-17
Adds chain management.

MicahZoltu:
Enables ethrpc to manage the blockchain client-side, rather than relying on the Ethereum node to do block chain management.  This gives much stronger guarantees that during a reorg or period of connectivity loss you will get notifications for all added and removed blocks.  This is a precursor to also getting strong guarantees of well ordered log message additions/removals.

### DIFF
--- a/src/block-management/block-stream.js
+++ b/src/block-management/block-stream.js
@@ -1,0 +1,169 @@
+"use strict";
+
+var BlockNotifier = require("./block-notifier.js");
+var Notifier = require("./notifier.js");
+var validateBlock = require("./validate-block.js");
+
+/**
+ * This object will emit a stream of new blocks, after handling backfills and reorgs.
+ * 
+ * @typedef transport
+ * @type object
+ * @property {function(string, boolean, function(object):void):void} getBlockByHash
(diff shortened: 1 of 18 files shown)


MicahZoltu:
@tinybike When you get a chance, can you re-teach coveralls to not comment on PRs?

src/block-management/block-stream.js
```diff
@@ -0,0 +1,169 @@
...
+
+  this.destroy = function () {
+    this.unsubscribeAll();
+    blockNotifier.destroy();
+  }.bind(this);
+
+  var shouldKeepBlock = function(block) {
+    if (!block) return true;
+    var blockNumber = parseInt(block.number);
+    if (isNaN(blockNumber)) return true;
```
kevinday: Would this only be true in the event of bad data? Why is the failure case to keep it?

MicahZoltu: Good catch, those early returns should both be `false`.  I think I originally had the function `shouldNotKeepBlock` and when I inverted it I missed those two early return statements.

PR #19, opened 2017-03-18
Replaces block subscriptions in ethrpc with new block stream.

MicahZoltu:
Tweaked the way transport reconnect subscriptions worked since the old way was lame (poor abstraction) and complicated integration with the new block stream.

Updates `ethereumjs-stub-rpc-server` which contains a bugfix that prevented the tests from passing.

### DIFF
--- a/src/transport/ipc-transport.js
+++ b/src/transport/ipc-transport.js
@@ -53,7 +53,7 @@ IpcTransport.prototype.submitRpcRequest = function (rpcJso, errorCallback) {
     });
   } catch (error) {
     if (error.code === "EPIPE") error.retryable = true;
-    errorCallback(error);
+    setTimeout(function() { errorCallback(error); });
   }
 };
 
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,7 +22,7 @@
             "internalConsoleOptions": "openOnSessionStart"
         },
         {
-            // docker run --rm -p 9229:9229 --name=ethrpc ethrpc --debug-brk --inspect --colors test/stub-node-only.js
+            // docker run --rm -p 9229:9229 --name=ethrpc ethrpc --debug-brk --inspect test/stub-node-only.js
             "type": "node2",
             "request": "attach",
             "name": "Attach to Docker",
--- a/test/stub-node-only.js
+++ b/test/stub-node-only.js
@@ -57,7 +57,8 @@ describe("tests that only work against stub server", function () {
           });
         });
 
-        it("starts connected > uses connection > loses connection > uses connection > reconnects > uses connection", function (done) {
+        // NOTE: this test is brittle on Linux.  see: https://github.com/nodejs/node/issues/11918
+        it("starts connected > uses connection > loses connection > uses connection > reconnects > uses connection (brittle)", function (done) {
           this.timeout(6000);
           stubRpcServer.addResponder(function (request) { if (request.method === "net_version") return "apple" });
           helpers.rpcConnect(transportType, transportAddress, function () {


src/index.js
```diff
@@ -149,7 +151,9 @@ module.exports = {
       // ensure we can do basic JSON-RPC over this connection
       this.version(function (errorOrResult) {
         if (errorOrResult instanceof Error || errorOrResult.error) return initialConnectCallback(errorOrResult);
-        this.setupBlockSubscription(function () { initialConnectCallback(null); });
+        this.blockStream = new BlockStream(createBlockStreamTransportAdapter(this), this.BLOCK_POLL_INTERVAL);
+        this.blockStream.subscribe(this.onNewBlock.bind(this));
+        initialConnectCallback(null);
```
MicahZoltu: There is a small (I believe) behavior change here where the connect callback will be called *before* the first block for the subscription has been received.  If this is undesirable, we could change `this.version` a few lines up to `this.getBlockByNumber('latest', false, this.onNextBlock)`.  I didn't do this because I don't think anyone actually cares about blocks right now, and users who care about blocks should call `ethrpc.blockStream.subscribe(callback)` rather than watching the blocks on ethrpc.

src/transport/ipc-transport.js
```diff
@@ -53,7 +53,7 @@ IpcTransport.prototype.submitRpcRequest = function (rpcJso, errorCallback) {
     });
   } catch (error) {
     if (error.code === "EPIPE") error.retryable = true;
-    errorCallback(error);
+    setTimeout(function() { errorCallback(error); });
```
MicahZoltu: This is just cleanup to make the sync error codepath functionally the same as the async error code path.

MicahZoltu:
No new tests were adding since this is an internal refactor, so existing coverage would have caught anything (and it did in the process of making this PR).

package.json
```diff
@@ -45,7 +45,7 @@
     "coveralls": "^2.11.4",
     "del": "^1.2.1",
     "ethereumjs-abi": "0.6.4",
-    "ethereumjs-stub-rpc-server": "1.0.0",
+    "ethereumjs-stub-rpc-server": "2.0.5",
```
MicahZoltu: There is a bug in the stub server that caused the tests to fail with this change.

src/block-management/ethrpc-transport-adapter.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = function (ethrpc) {
+  return {
+    getBlockByNumber: ethrpc.getBlockByNumber.bind(ethrpc),
+    getBlockByHash: ethrpc.getBlockByHash.bind(ethrpc),
+    subscribeToReconnects: ethrpc.internalState.transporter.addReconnectListener.bind(ethrpc.internalState.transporter),
+    unsubscribeFromReconnects: ethrpc.internalState.transporter.removeReconnectListener.bind(ethrpc.internalState.transporter),
+    subscribeToNewHeads: function (onNewHead, onSubscriptionError) {
+      var token = (nextToken++).toString();
+      subscriptionMapping[token] = null;
+      try {
```
tinybike: What errors can be picked up by this try/catch?  Since all that's inside the try is an async function and its callback, would it make more sense to put the try/catch inside the callback (possibly w/ catch passing the exception to `onSubscriptionError`)?

src/transport/abstract-transport.js
```diff
@@ -20,6 +20,7 @@ function AbstractTransport(address, timeout, messageHandler) {
   this.awaitingPump = false;
   this.connected = false;
   this.backoffMilliseconds = 1;
+  this.nextReconnectListenerToken = 1;
```
tinybike: Can you explain how multiple reconnect tokens are used here?  It seems (naively?) possible to have just a single block subscription property, which is set to the subscription ID, and is replaced any time reconnect happens.  When would it be necessary to track multiple newHeads subscriptions?

src/block-management/ethrpc-transport-adapter.js
MicahZoltu: The try/catch is defensive against external (not part of this class) code throwing for some reason.  However, you are right that it is inconsistent with the rest of this project (which isn't written defensively) and even the rest of the code in this PR and the previous, so I have removed it.

src/block-management/ethrpc-transport-adapter.js
```diff
@@ -0,0 +1,41 @@
+"use strict";
+
+var nextToken = 1;
```
tinybike: Is this the same token that is attached to AbstractTransport as `nextReconnectListenerToken`?  If so, it seems confusing (to me) to track and increment them separately.  If not, I'm not clear on the purposes of the two sorts of tokens -- could you explain a bit?

Related thought: since the tokens are just integers that are incremented by one each time, could `subscriptionMapping` be replaced with an array (and nextToken/nextReconnectListenerToken simply be the length of the array)?

src/transport/abstract-transport.js
MicahZoltu: I'm not sure I fully understand the question.  This is the auto-incrementing (guaranteed unique) token returned from calls to `<some_transport>.addReconnectListener`.  It can be used by the caller to stop getting notified about reconnects with `<some_transport>.removeReconnectListener`.  This has nothing to do with blocks or block streams (though it is part of the foundation upon which reliable block streams are built).

tinybike: Ah, alright, so this is a completely different token than the one tracked by ethrpc-transport-adapter.  Does this question make more sense applied to the `nextToken` variable in ethrpc-transport-adapter?

src/block-management/ethrpc-transport-adapter.js
MicahZoltu: This is not the same token that is inside `AbstractTransport`.  You will find that this pattern is followed a lot in the various "classes" that were created in this PR, the previous one, and the upcoming one.  The tokens are not interchangeable in any way and each stands on its own.  In all cases, they are used to allow the caller to unsubscribe from something he previously subscribed to.  Reconnects, blocks, logs, etc.  If a function like `foo.subscribe` returns a token, that token is only useful for calling `foo.unsubscribe`.  It is not useful for calling `bar.unsubscribe` or `zip.removeReconnectListener`.

> could subscriptionMapping be replaced with an array

This would lead to a sparse array over time if someone is connecting/reconnecting which isn't good for JavaScript performance (though pragmatically it probably wouldn't matter much at the moment).

> since the tokens are just integers

The returned tokens are explicitly strings not numbers.  The fact that they happen to be auto-incremented numbers is an implementation detail and subject to change.  A UUI with class prefix would probably be a better system for generating tokens as it would allow the unsubscribe methods to validate that the token provided was reasonable, but I didn't feel writing the code associated with or bringing in the dependencies required to make that happen.

src/transport/abstract-transport.js
MicahZoltu: > When would it be necessary to track multiple newHeads subscriptions?

Answering this question in isolation (ignoring context of the rest of the comment):
`BlockStream` is built generically without making assumptions on how it will will be used.  In theory, a single `BlockStream` would be instantiated at some point but multiple callers can subscribe to notifications of new blocks.  This is useful if different parts of code want notifications independently.  One or more of those may be some kind of short-lived thing (e.g., a UI dialog that may be closed by the user) and as such they would want to unsubscribe when they are done so their code stops running and can get garbage collected.  We don't want to destroy the entire `BlockStream` object every time someone wants to unsubscribe from new block notification, so we allow as many callers to subscribe as want, and we give them tokens back so they can later unsubscribe (if they want).

If you are referring to the actual block subscription to the remote Ethereum Node (i.e., `eth_subscribe`) then there will be one per instance of `BlockStream`.  In most cases, the app will probably only have one `BlockStream` for its life and it can be treated as a singleton.  However, the code is written so this isn't strictly necessary (makes testing easier and could be useful one day).

src/block-management/ethrpc-transport-adapter.js
tinybike: > The returned tokens are explicitly strings not numbers.
Why is this?  JS will auto-convert integer keys to strings -- I guess you just prefer to be explicit?

> The fact that they happen to be auto-incremented numbers is an implementation detail and subject to change. A UUI with class prefix would probably be a better system for generating tokens as it would allow the unsubscribe methods to validate that the token provided was reasonable, but I didn't feel writing the code associated with or bringing in the dependencies required to make that happen.

If you intend to change it, IMO the time to do that is now, while all this is still fluid and fresh in your mind.  If you want to use UUIDs specifically, the `uuid` and `uuid-parse` libraries are already dependencies of augur.js -- feel free to bring them into ethrpc.

src/transport/abstract-transport.js
tinybike: > If you are referring to the actual block subscription to the remote Ethereum Node (i.e., `eth_subscribe`) then there will be one per instance of `BlockStream`.  In most cases, the app will probably only have one `BlockStream` for its life and it can be treated as a singleton.  However, the code is written so this isn't strictly necessary (makes testing easier and could be useful one day).

When you instantiate `BlockStream`, this automatically instantiates a `BlockNotifier`, which instantiates either a PollingBlockNotifier or a SubscriptionBlockNotifier, which in turn automatically calls `transport.subscribeToNewHeads` which -- if I'm following this correctly -- is the function defined in ethrpc-transport-adapter that increments `nextToken` and also calls `ethrpc.subscribeNewHeads` (which calls `eth_subscribe` under the hood).  Is this correct, or have I missed something?  If it's correct, then `nextToken` should only get incremented when a new `eth_subscribe` is issued.

tinybike:
A few comments / issues:

1. In general, it's better to attach methods to the prototype rather than to the instance in the constructor, since if they're attached to the instance the methods get re-declared for every instance (which can be expensive if there are many objects created). Ex: `this.prototype.myFunc = function () { ... }` is usually preferable to `this.myFunc = function () { ... }.bind(this)` inside the constructor.

2. If you're only binding `this` for use inside a closure, `var self = this` in the enclosing scope is preferred over `myFunc = function () {}.bind(this)` as it is more performant and is the convention used throughout the codebase.

3. Please don't use the following pattern to create "private" variables and methods:
```js
function MyConstructor() {
    var myPrivateVar;
    var myPrivateFunction = function () {};
}
```
IMO this is an anti-pattern, as it makes these "private" values and methods untestable.  In general, I believe that dealing with complexity by hiding it is a bad idea.  (I'm of course aware that this is "vanilla OO" stuff, but augur's stack is not written in OO style, which is very much intentional.)

4. Have you looked at using Node's built-in `stream` module for `BlockStream`?  Unless it's incompatible for some reason, it's probably a good idea to use it, as it is what third-party consumers will expect in a streaming API (assuming ethrpc ever has third-party users).

src/transport/abstract-transport.js
MicahZoltu: _Assuming_ all of this code is only ever used in this one way and nothing ever changes over time, then I believe you are correct that `ethrpc.subscribeNewHeads` will only be called once per instantiation of `BlockStream` and there will likely only be one token ever used.

MicahZoltu:
> In general, it's better to attach methods to the prototype rather than to the instance in the constructor, since if they're attached to the instance the methods get re-declared for every instance (which can be expensive if there are many objects created). Ex: this.prototype.myFunc = function () { ... } is usually preferable to this.myFunc = function () { ... }.bind(this) inside the constructor.

These are long lived objects, not short lived ones with high creation frequency so the cost of their instantiation is not significant with regards to performance.  Also, I believe the only performance difference is memory, which is negligible for anything but objects created in very high-volume.  The reason they are not on the prototype is so they have access to the the constructor closure, which is a separate topic of discussion (below).

> If you're only binding this for use inside a closure, var self = this in the enclosing scope is preferred over myFunc = function () {}.bind(this) as it is more performant and is the convention used throughout the codebase.

I believe under the hood using a `self` variable and `.bind(this)` are pragmatically the same.  One could argue that `.bind` creates a wrapping closure in some JS engines that is technically more costly (extra function call) but this is negligible relative to basically everything.  The reason I prefer `.bind(this)` is because it is a pattern I can liberally follow everywhere without any thought.  With `var self = this` you have to capture `self` while you have a valid `this` (and hope that all entry points have `this` set correctly) and then use `.bind(this)` anytime your entrypoints may be used in a way where `this` isn't set to the expected value (e.g., using the first function as a callback).  This pattern (`bind(this)`) is also effectively what ES6 went with to deal with the "`this` problem" in ES6 in that all `() => {}` functions auto-bind(this).

That being said, your last argument rings true with me: "[self = this] is the convention used throughout the codebase." so I can switch patterns for consistency.

> it makes these "private" values and methods untestable.

That is part of the point.  😄  You should not be testing internal implementation details of a data structure, you should be testing the public surface area.  A perceived need to test implementation details is a code smell that usually indicates problems with abstraction, encapsulation, layout, etc.  By hiding these functions, it prevents me and future authors from testing something that shouldn't be tested, thereby making refactoring difficult.

> this is an anti-pattern [...] I believe that dealing with complexity by hiding it is a bad idea.

Privacy allows future readers to understand what the public surface area of a collection of code is without having to grok the whole thing.  It would allow someone to look at BlockStream (or any file/class) and, without much effort, identify exactly how they are intended to interact with it.  It also makes it clear to the tester what should be tested and what shouldn't, and the code reviewer what usage patterns to expect.  I am curious why you consider privacy/data-hiding an anti-pattern?  Privacy isn't just an OO concept, it is also found in functional languages in a different form (e.g., nested functions, closures, etc.).

----

The reason I have taken on most of these patterns that you have brought up here is because I see these code-bases (ethrpc, ethereumjs-connect, augur.js, etc.) as long-lived code-bases, not throw-away code.  When you have a long-lived code-base, there is great value in authoring it so that the encapsulations/abstractions are very clear to future readers and intent is well documented in code.  When this project sits untouched for 6-months or someone new comes on to work with it, that person will not have the context that we have today with regards to how things are supposed to be used, how things are expected to interact, etc.  By making it so the code can only be used in the way it was intended at design-time, we are future proofing the code-base against future maintainers (including you/me after we have forgotten the intricacies of how it is supposed to work).

----

Of course, as always, this is your code-base so I'll make all of the changes you request.  😀

tinybike:
>  I am curious why you consider privacy/data-hiding an anti-pattern?

Because it prevents thorough testing (see below).

> That is part of the point.  :smile: You should not be testing internal implementation details of a data structure, you should be testing the public surface area. A perceived need to test implementation details is a code smell that usually indicates problems with abstraction, encapsulation, layout, etc. By hiding these functions, it prevents me and future authors from testing something that shouldn't be tested, thereby making refactoring difficult.

In our codebase, we have both integration tests, which are basically tests of the public surface area, as well as unit tests, which are tests of individual functions.  Unit tests are expected to test internal implementation details.  You may regard this as a "code smell", but for augur's middleware, it was an explicit design decision to structure the tests this way -- augur.js, keythereum, etc. have unit tests at the level of individual functions.  Every single function in the middleware should be exhaustively tested in this way, regardless of whether it is part of the intended public API.  (Where coverage is incomplete, it's because we simply haven't gotten around to adding thorough tests, not because we intend to leave them that way.)

ethrpc will need to be exhaustively unit tested in the same manner as the other middleware components.  This includes your new code!  In my experience, this sort of rigorous testing catches quite a few bugs, and results in more reliable software.  I understand that you disagree that this is necessary/useful, but it is a requirement for this project.

(I know requirements like this should really be laid out up-front, rather than dragged in piecemeal as part of code reviews.  The JS style guide I'm putting together should hopefully make this all more cut-and-dried.)

MicahZoltu:
> In our codebase, we have both integration tests, which are basically tests of the public surface area, as well as unit tests, which are tests of individual functions. Unit tests are expected to test internal implementation details.

In any given application, there are many layers of "public surface area".  There is the public surface area of the library as a whole (this aligns with your definition of public surface area above) and there is the public surface area to a file (what is exported via module.exports) and then there is the public surface area of a class (its this.* and its prototype for JavaScript) and finally there is the public surface area of a function (its signature).  When it comes to unit testing, the goal of a unit test is to test the smallest _interesting_ _unit_ of work.  Taken to the extreme, one could argue that a single CPU instruction is the smallest unit of work, but this isn't interesting nor useful to test (unless you are building CPUs).  Unwinding a bit, one could argue that the smallest unit of interesting work is a single line of a high level language, but again this isn't really "interesting" to the user.  Generally speaking, when people talk about unit tests the "unit" they are referring to tends to be the *largest* fully encapsulated piece of work in a system.  In OO languages this usually means a class (but not its private parts) and in functional languages this usually means an exposed function (but not its locals, captures and nested functions).  The reason for this distinction is because unit testing of internal implementation details (and by internal I mean internal to the "unit" as defined above) makes the code *less* agile and *less* maintainable because touching *anything* breaks it.  It also makes it so your tests aren't actually telling you anything other than "someone wrote these".

For an extreme example, if one were to have unit tests that ensured that the 4th line of a particular function changed the state of some local variable to 5, then this function can't be touched/changed in any way without breaking the test.  This test doesn't provide any meaningful value to anyone because no one *cares* that the 4th line of that function sets a local variable to 5.  They *care* that the function does what it says it does.  What this test *does* do is make it so if someone wants to change this function in any way (even to fix a bug on an unrelated line above line 4) they need to also fix a test that isn't testing anything meaningful.  This, in turn, leads to test failure apathy where people just assume that any change will break tests and stop looking at the test failures as an indicator that the functionality may have changed in a _meaningful_ way.

For a more pragmatic example, say you have a function that adds 1 and 2 and returns the result.

Initial implementation:
```javascript
function _addOne(initial) {
	var result = initial;
	for (var i = 0; i < 1; ++i) {
		++result;
	}
	return result;
}
function _addTwo(initial) {
	var result = initial;
	for (var i = 0; i < 2; ++i) {
		++result;
	}
	return result;
}
function addOneAndTwo() {
	var result = 0;
	result = _addOne(result);
	result = _addTwo(result);
	return result;
}
```

Refactored implementation:
```javascript
function addOneAndTwo() {
	return 3;
}
```

The only function that we care about for this unit of work (say a file in this case) is `addOneAndTwo`.  We don't ever use `_addOne` or `_addTwo` other than as an implementation detail to how `addOneAndTwo` works.  They aren't intended to be used outside of this file and we aren't making any promises that they'll stick around.  In fact, we could have just as easily inlined the body of `_addOne` and `_addTwo` but we extracted them to make the code more readable.  If we were to write tests of `_addOne` and `_addTwo` then when we came in to refactor our code (make changes to how it is implemented without any expectation of external behavior changing) the tests would break, even though external behavior did not change.  As mentioned above, this results in developers learning to ignore failing tests as "meh, we refactored the code so tests were expected to break" and they juts go and change the tests.  This is problematic because this developer no longer can tell the difference between expected external effects and internal implementation details.

----

> ethrpc will need to be exhaustively unit tested [...]. This includes your new code!

Given the above definition of "unit", all of the new code is thoroughly unit tested.  All expected external behaviors are fully tested (as far as I know).  Implementation details of the various units of work are *not* tested, because that would not add value (it would remove value).  I fairly strongly believe that the testing I have done here is *more* robust than the testing you have proposed.

----

> this sort of rigorous testing catches quite a few bugs, and results in more reliable software

The testing you have described actually does the opposite in reality.  Don't get me wrong, I'm not disagreeing with the testing philosophy at all, but implemented the way you have proposed will result in less agility, less readable/maintainable code, and likely more bugs due to developer apathy and unclear boundaries.

----

To look at this another way, I could have implemented 98% of the `BlockStream` entirely within a single function (`processMaybeNewBlock`).  I _chose_ to break it apart to make the code easier to understand, more readable, etc.  If there is an expectation to test every one of those breakout functions, I would have been encouraged to instead write it as a single mega-function that is less readable.  If I make all of those internal helper methods externally visible and then write tests for them all, not only would the usage pattern of BlockStream become far less understandable, I would be unable to refactor *how* `BlockStream` does what it does without having to rewrite almost all of the tests.  I also wouldn't be able to tell the difference between tests that verify expected external functionality and tests that just cover some random helper function (which could be removed, renamed, changed, etc. without effecting real functionality).

MicahZoltu:
Closing this as we now have `ethereumjs/ethereumjs-blockstream` to do most of the work done in this PR.  I'll be opening another PR that leverages that instead.

PR #20, opened 2017-03-30
Some minor cleanup work.

MicahZoltu:


MicahZoltu:
This is just some minor fixes/cleanup.  Going to merge as soon as the checks pass unless someone speaks up before that.  If any of this is found to be contentious, I'll roll it back.

src/transport/ipc-transport.js
```diff
@@ -53,7 +53,7 @@ IpcTransport.prototype.submitRpcRequest = function (rpcJso, errorCallback) {
     });
   } catch (error) {
     if (error.code === "EPIPE") error.retryable = true;
-    errorCallback(error);
+    setTimeout(function() { errorCallback(error); });
```
tinybike: What's the setTimeout doing here?

MicahZoltu: It forces the error callback to be called in an asynchronous context rather than a synchronous context.  When a user calls `submitRpcRequest`, if the call fails asynchronously the callback will be fired in an async context *after* `submitRpcRequest` returns.  However, if `ipcClient.write` throws before it switches contexts then the error will occur in the synchronous context.

By adding a `setTimeout` around `errorCallback`, we guarantee that `errorCallback` will always be called in an async context rather than sometimes async and sometimes sync.

While I generally do this as a best-practice, in this case it actually matters because the error callback may result in a retry, which may result in another failure and too many of those in succession could result in a stack overflow.  By forcing the error callback into an async context we guarantee that the stack depth will remain small no matter how many times the process is repeated.

tinybike: Interesting, that makes sense.  When I've used `setTimeout` to force something to execute asynchronously, I always use `setTimeout(() => {/* hilarity ensues */}, 0)`.  I guess it accomplishes the same thing if you just leave out the 2nd parameter though?

MicahZoltu: Yeah, `setTimeout` with no duration is the same as `0` duration which effectively means that the timeout is elligible for execution immediately, which means the next time timeouts are executed, this one will be executed.

PR #24, opened 2017-04-04
Fixes a bug on connect where it wasn't falling back to reasonable defaults.

MicahZoltu:
This is pretty severe as an `undefined` `pollingIntervalMilliseconds` results in the block poller spamming the node for new blocks (effectively `setTimeout(callback, 0)`).


MicahZoltu:
As this is a pretty severe bug (causes chrome to peg your CPU) and the fix is "safe", I'm going to merge as soon as the checks pass.  I'll revert if anyone has a problem with the changes.

PR #26, opened 2017-04-14
Reduxification / pre-audit cleanup

tinybike:
@MicahZoltu and @stephensprinkle: could you guys review this PR?  Summary: I moved some stuff around in ethrpc and changed the way it deals with state (it uses Redux now).

Micah: this PR moves all of the main ethrpc state (public and internal) to the Redux store.  It does *not* change anything about the internal states maintained by Transporter or BlockNotifier instances.  When you're reviewing this PR, please think about the best way to move Transporter and BlockNotifier over to this new model!

One issue I'm aware of is that the Redux store is currently storing several functions -- when users register callbacks, those have to be saved somewhere, so they are right now put in the store with the other data.  Possibly this can be done more cleanly using the store's `subscribe` method?


src/block-management/on-new-block.js
```diff
@@ -0,0 +1,19 @@
...
+var assign = require("lodash.assign");
+var reprocessTransactions = require("../transact/reprocess-transactions");
+var isObject = require("../utils/is-object");
+
+function onNewBlock(block) {
+  return function (dispatch) {
+    if (isObject(block)) {
+      dispatch({
+        type: "SET_CURRENT_BLOCK",
+        block: assign({}, block, { number: parseInt(block.number, 16) })
```
MicahZoltu: Since this PR will be a semver major change, consider taking the opportunity to no longer do this mutation on the block and instead store the raw block?  Having one field on the block be not in the native format can be confusing for users.  Alternatively, we could create a "ParsedBlock" that turns everything into byte arrays/BigNumbers so the stored block is easier to work with.

src/decode-response/convert-response-to-returns-type.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
+
+var abi = require("augur-abi");
+var clone = require("clone");
+
+var convertResponseToReturnsType = function (returnsType, response) {
+  var convertedResponse;
+  if (!returnsType) return response;
+  if (response && response !== "0x") {
```
MicahZoltu: Consider inverting this if into a couple guard clauses:
```
if (!response) return response;
if (response === "0x") return response;
if (response.error) return response;
...
```
This will reduce nesting by a level and make the function easier to read (lower cyclomatic complexity).

src/decode-response/convert-response-to-returns-type.js
```diff
@@ -0,0 +1,39 @@
...
+      convertedResponse = abi.unfix_signed(convertedResponse, "string");
+    } else if (returnsType === "null") {
+      convertedResponse = null;
+    } else if (returnsType === "address" || returnsType === "address[]") {
+      convertedResponse = abi.format_address(convertedResponse);
+    }
+  } else {
+    convertedResponse = response;
+  }
+  return convertedResponse;
```
MicahZoltu: Instead of setting `convertedResponse` all throughout the function, consider just directly returning once you have a result.  This will simplify the function and allow the reader to know when the function is done when reading without having to scan for further changes to `convertedResponse`.

src/decode-response/handle-rpc-error.js
```diff
@@ -0,0 +1,34 @@
+"use strict";
+
+var abi = require("augur-abi");
+var errors = require("../errors/codes");
+
+var handleRPCError = function (method, returns, response) {
+  var i, len, responseNumber;
+  if (response) {
```
MicahZoltu: Same as above, recommend inverting the if into a guard clause:
```
if (!response) return response;
...
```
Reduces complexity and makes the code easier to read.  Same applies to all other places where an `if (...) return` could avoid nesting, I'll not comment on the rest though.

src/block-management/create-block-and-log-streamer.js
```diff
@@ -0,0 +1,58 @@
...
+ * @property {function(function():void, function(Error):void):string} subscribeToReconnects
+ * @property {function(string):void} unsubscribeFromReconnects
+ * @property {function(function(Block):void, function(Error):void):string} subscribeToNewHeads
+ * @property {function(string):void} unsubscribeFromNewHeads
+ *
+ * @param {Configuration} configuration
+ * @param {Transport} transport
+ */
+function createBlockAndLogStreamer(configuration, transport) {
+  return function (dispatch) {
```
MicahZoltu: Consider binding at call site instead of returning a function:
```
function createBlockAndLogStreamer(configuration, transport, dispatch) {
	// no inner function return here
	var reconcileWithErrorLogging, blockNotifier, blockAndLogStreamer;
	...
}
store.dispatch(connect.bind(configuration, initialConnectCallback));
````

Same suggestion for all others.

File: src/index.js [hunk 6]
```diff
@@ -5,1981 +5,235 @@
...
+  eth: bindDispatch(dispatch, eth),
+  net: bindDispatch(dispatch, net_),
+  web3: bindDispatch(dispatch, web3),
+  txpool: bindDispatch(dispatch, txpool),
+  shh: bindDispatch(dispatch, shh),
+  miner: bindDispatch(dispatch, miner),
+  personal: bindDispatch(dispatch, personal),
+
+  // web3_
+  sha3: sha3,
```
MicahZoltu: Why isn't this namespaced like the others?  (same question for the below)

src/platform/request.js
```diff
@@ -1,7 +1,4 @@
 "use strict";
 
 var isNodeJs = require("./is-node-js.js");
-if (isNodeJs)
-	module.exports = require("request");
-else
-	module.exports = require("browser-request");
+if (isNodeJs)  {module.exports = require("request");} else	{module.exports = require("browser-request");}
```
MicahZoltu: Extraneous whitespace here (tab instead of space)

MicahZoltu: Out of scope of PR, but this should really be a ternary:
```
module.exports = isNodeJs ? require("request") : require("browser-request");
```

src/reducers/configuration.js
```diff
@@ -0,0 +1,40 @@
...
+  errorHandler: null
+};
+
+module.exports = function (configuration, action) {
+  var updatedConfiguration;
+  if (typeof configuration === "undefined") {
+    return initialState;
+  }
+  switch (action.type) {
+    case "SET_CONFIGURATION":
```
MicahZoltu: If a user calls `SET_CONFIGURATION` not all of their configuration changes will occur.  The reason configuration was passed into `connect` previously was to ensure that configuration changes were applied everyone and to not give the user an impression that they could set configuration later and expect it to have an effect.

Recommend not storing configuration directly in Redux or updating this reducer to have side effects of resetting everything (reconnect, etc.).  I'm not personally a fan of the latter.

src/block-management/ethrpc-transport-adapter.js
```diff
@@ -1,47 +1,61 @@
...
+          // if the caller already unsubscribed by the time this callback is
+          // called, we need to unsubscribe from the remote
+          if (subscriptionMapping[token] === undefined) {
+            dispatch(eth.unsubscribe(subscriptionID, noop));
+          } else {
+            subscriptionMapping[token] = subscriptionID;
+            dispatch(addNewHeadsSubscription(subscriptionID, onNewHead));
+          }
+        }));
+        return "1";
```
MicahZoltu: Why is this returning `"1"` instead of `token` now?  The token is used to unsubscribe from notifications (nothing more).  Purely by nature of `token` being initialized to `0` will this work correctly for the first call to `subscribeToNewHeads`, but it will break for any subsequent calls.

src/block-management/ethrpc-transport-adapter.js
```diff
@@ -1,47 +1,61 @@
...
+        return transporter.addReconnectListener(onReconnect);
+      },
+      unsubscribeFromReconnects: function (token) {
+        transporter.removeReconnectListener(token);
+      },
+      subscribeToNewHeads: function (onNewHead, onSubscriptionError) {
+        var token = (nextToken++).toString();
+        subscriptionMapping[token] = null;
+        dispatch(eth.subscribe(["newHeads", {}], function (subscriptionID) {
+          if (!subscriptionID || subscriptionID.error) {
```
MicahZoltu: Are you certain that `subscriptionID` will never be an `Error` object?

src/block-management/ethrpc-transport-adapter.js
tinybike: Weird, that's a typo, not sure how that happened.

src/block-management/ethrpc-transport-adapter.js
tinybike: Nothing coming back from the blockchain should be an `Error` object.  Did any of your transport changes make it so an error object could appear in an async RPC callback?

src/reducers/configuration.js
tinybike: Users aren't expected to call the `SET_CONFIGURATION` action itself, though.  The raw actions are technically accessible, but they're not part of the normal public API.

I think storing configuration in Redux makes sense (minus the out-of-band error handler, which is now stored elsewhere).  It's simple, serializable, and quite relevant to ethrpc's behavior -- if you were going to rehydrate and/or rewind, it seems to me you'd lose an awful lot of context if you didn't know (for example) what endpoints were configured at the time.

src/platform/request.js
tinybike: Yikes, eslint's autoformatter did something bad here!

src/index.js
tinybike: It's because `sha3` has optimized by doing the calculation locally, so I see no real use-case for the actual `web3_sha3` RPC.

src/decode-response/handle-rpc-error.js
tinybike: Googling a bit reveals that the holy war between early-returns and single-point-of-exit rages on.  I try and be pragmatic about it.  I think I agree with you in this case though that an early return would be cleaner.

tinybike: Are there any cases where non-batched RPC requests result in an array coming back from the node?  If not there are a number of `if array do X else do Y` conditionals that we could chop out of the code...

src/wrappers/make-wrapper.js
```diff
@@ -0,0 +1,17 @@
...
+var raw = require("./raw");
+var isFunction = require("../utils/is-function");
+
+function makeWrapper(command) {
+  return function (params, callback) {
+    return function (dispatch) {
+      if (callback === undefined && isFunction(params)) {
+        return dispatch(raw(command, [], params));
+      }
+      return dispatch(raw(command, params, callback));
```
tinybike: The signature is the same for all of these named wrappers though -- it's always just `namespace_methodName(params, callback)`.  The reason I like the named wrappers for internal use is because they let you declare which namespace(s) you're using when you're requiring.  If I `require('../wrappers/raw')` I could send anything at all through `raw`.

tinybike:
Re: the larger point about putting non-serializable things into Redux.  The callbacks we're storing are for the most part not really "state" per se, but rather state observers.  Because they aren't serializable, if you try to persist-and-rehydrate (e.g. from localStorage) your rehydrated ethrpc won't have the actual *function* used for the callback -- instead it'll just have a record that there was a callback stored there.  Same with time-travel: rewind-and-reload will include the info that a callback was placed into the store, but not the callback itself.

I think this issue ultimately stems from the fact that library code is "incomplete", in the sense that there can be user-supplied functions that are not known to the authors of the library.  Here are some [further thoughts](https://github.com/tinybike/redux-thunk-subscribe#motivation) on this issue.  My conclusion is:
1. For callbacks that are *not* store observers, it doesn't particularly matter where they're stored -- so for the sake of a more convenient workflow I've taken these out of the Redux store and they are now kept in "conventional" state.
2. Callbacks that *are* store observers should simply `subscribe` to the store, so that they can react to the relevant state changes.  (The whole point of this is, after all, to program in a reactive style!)  To make this more convenient, I wrote [Redux Thunk Subscribe](https://github.com/tinybike/redux-thunk-subscribe), which makes `store.subscribe` available directly in the thunks' signatures, and ethrpc now uses this in place of the [Redux Thunk](https://github.com/gaearon/redux-thunk) middleware.

tinybike:
Addendum: the reason I want to place some of ethrpc's state (the good parts!) into Redux is so that UIs (in particular ours) can react to changes in the middleware in the same way they do to changes in the UIs's own Redux store.  e.g., ethrpc saves a bunch of information about blocks and transactions in its internal state.  But, those changes can't automatically propagate to React (the view layer) because React can't subscribe to some random object floating around in the middleware.

My big-picture goal here / why I'm going through this spasm of intense edits to ethrpc is to allow consumers of ethrpc to *pass in their own Redux store*, so that the consumers can react to changes in ethrpc's Reduxified state.

tinybike:
Also, augur.js is darn close to being stateless.  Stateless augur.js + stateless ethereumjs-connect + ethrpc and UI Reduxified and using the same store = reactive as *HELL*.

src/decode-response/handle-rpc-error.js
MicahZoltu: Yeah, `getLogs` returns an array of logs (just one example).

src/block-management/ethrpc-transport-adapter.js
MicahZoltu: I believe everything goes through `parse`, so whatever `parse` does.  Glancing at it, I see one code path where it is blindly passing `response.error` but otherwise it appears that it manually constructs error objects for all responses.

PR #27, opened 2017-04-17
Adds support for signing with a function instead of a key.

MicahZoltu:
This enables hooking up hardware signers at a higher layer.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1250,8 +1250,18 @@ module.exports = {
    * @param {buffer} privateKey The sender's plaintext private key.
    * @return {string} Signed and serialized raw transaction.
    */
-  signRawTransaction: function (packaged, privateKey) {
-    var rawTransaction = new EthTx(packaged);
+  signRawTransaction: function (packaged, privateKeyOrSigner, callback) {
+    try {
+      var transaction = new EthTx(packaged);
+      return (isFunction(privateKeyOrSigner))
+        ? privateKeyOrSigner(transaction, callback)
+        : this.signTransactionWithKey(transaction, privateKeyOrSigner, callback);
+    } catch (error) {
+      return callback(error, undefined);
+    }
+  },
+
+  signTransactionWithKey: function (rawTransaction, privateKey) {
     rawTransaction.sign(privateKey);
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -1250,8 +1250,18 @@ module.exports = {
...
-  signRawTransaction: function (packaged, privateKey) {
-    var rawTransaction = new EthTx(packaged);
+  signRawTransaction: function (packaged, privateKeyOrSigner, callback) {
+    try {
+      var transaction = new EthTx(packaged);
+      return (typeof privateKeyOrSigner === "function")
+        ? privateKeyOrSigner(transaction, callback)
+        : this.signTransactionWithKey(transaction, privateKeyOrSigner, callback);
+    } catch (error) {
+      return callback(error, undefined);
```
tinybike: While I personally like `callback(error, result)`, I recommend using the `callback(error || result)` style here for consistency.

src/index.js
```diff
@@ -1392,21 +1402,15 @@ module.exports = {
       console.log("[ethrpc] packaged:", JSON.stringify(packaged, null, 2));
     }
     if (!isFunction(callback)) {
-      return this.signRawTransaction(
-        this.setRawTransactionNonce(this.setRawTransactionGasPrice(packaged), address),
-        privateKey
-      );
+      if (typeof privateKeyOrSigner === "function") throw new Error("Cannot do a synchronous signing with a signer function.");
```
tinybike: Recommend using `isFunction(privateKeyOrSigner)` for consistency.

tinybike:
Made a couple small suggestions.  Otherwise this looks good.

src/index.js
MicahZoltu: The problem with `(error || result)` is that while this library uses `(errror || result)`, the callback will be passed through to the user-provided signer function.  Since `(error, result)` is the more common signature throughout the JavaScript ecosystem it seems like in this case we should follow that instead of forcing all users of this library to conform to the `(error || result)` style that this library uses.

src/index.js
MicahZoltu: Fixed.

PR #28, opened 2017-04-25
Fix `rpc.connect` in README.md

fanatid:


MicahZoltu:
Changes look fine.  Something is wrong with TravisCI though.  This is readme only so definitely not the cause.

PR #30, opened 2017-04-26
Use `abi.number` instead `parseInt`

fanatid:
Is there any reason why `parseInt` used for `chainId` and `abi.number` for `gasPrice`?


tinybike:
No, there isn't.  I agree that it's best to be consistent.

PR #35, opened 2017-10-20
Remove Node 7 version from travis build

holgerd77:
I think it is not worth to fix Node 7 test run, since the version is [not supported](https://github.com/nodejs/Release) any more, so I removed it from the travis build file to get the build passing again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
 
 node_js:
   - "8"
-  - "7"
   - "6"
 
 addons:


MicahZoltu:
I'm 👍 on this, though it is ultimately @tinybike's call.

holgerd77:
@tinybike Could you drop a note on this?

tinybike:
I'm fine with removing Node 7!

PR #36, opened 2017-12-01
split connecting and blockstream listening

epheph:
@pgebheim this is your branch, with my commit on top to use `dispatch()`


epheph:
Current test failure is sporadic, not related to this change

PR #37, opened 2017-12-08
Double call is occurring upon error

canuc:
In the case there is a validation error callback is being called twice.


canuc:
fixes: https://github.com/ethereumjs/ethrpc/issues/38

Issue #32, opened 2017-05-12
Address validation


fanatid:
Address validation use regexp `^0x[0-9a-fA-F]*$` https://github.com/ethereumjs/ethrpc/blob/4f3789895d5ad7830befe629f7bb304afb28ecd0/src/validate/validate-address.js#L14
not sure, but should not you check checksum if address has upper case symbols?

MicahZoltu:
I believe better address validation could be done.  What is in there now is a pretty naive sanity check.  If you happen to have a link to a better mechanism for address validation feel free to drop a link.  :)

fanatid:
@MicahZoltu this should work:
```js
"use strict";
var sha3 = require("../utils/sha3");

function checksumAddress (address) {
  address = address.slice(2).toLowerCase();
  var hash = sha3(address, "hex");
  var ret = "0x";

  for (var i = 0; i < address.length; i++) {
    if (parseInt(hash[i], 16) >= 8) {
      ret += address[i].toUpperCase();
    } else {
      ret += address[i];
    }
  }

  return ret;
}

var validateAddress = function (address) {
  if (address === null || address === undefined) {
    throw new Error("address is required");
  }
  if (typeof address !== "string") {
    throw new Error("address must be a string but was " + typeof address);
  }
  if (address.slice(0, 2) !== "0x") {
    address = "0x" + address;
  }
  if (address.length !== 42) {
    throw new Error("address must be 42 characters, 20 bytes (2 hex encoded code points each) plus the 0x prefix.  Length: " + address.length);
  }
  if (!/^0x[0-9a-fA-F]{40}$/.test(address)) {
    throw new Error("address can only contain 0-9 and a-F and must start with 0x.  Provided: " + address);
  }
  if (/[a-f]+/.test(address) && /[A-F]+/.test(address) && checksumAddress(address) !== address) {
    throw new Error("address checksum is invalid.");
  }
  return address;
};

module.exports = validateAddress;
```

MicahZoltu:
@fanatid I'm guessing at this point you don't really care anymore, but I just noticed this issue was never implemented.  A PR is welcome.   Same to anyone else interested in this, a PR that improves the validation would be welcome (should be as simple as just dropping the above code into a file and hooking it up).

MicahZoltu:
Fixed by #42

kiter32:
[...]

PR #42, opened 2017-12-21
implement #32 suggestions

amousa11:
This code was copied from @fanatid in https://github.com/ethereumjs/ethrpc/issues/32

### DIFF
--- a/src/validate/validate-address.js
+++ b/src/validate/validate-address.js
@@ -1,4 +1,21 @@
 "use strict";
+const sha3 = require("../utils/sha3");
+
+function checksumAddress (address) {
+  address = address.slice(2).toLowerCase();
+  var hash = sha3(address, "hex");
+  var ret = "0x";
+
+  for (var i = 0; i < address.length; i++) {
+    if (parseInt(hash[i], 16) >= 8) {
+      ret += address[i].toUpperCase();
+    } else {
+      ret += address[i];
+    }
+  }
+
+  return ret;
+}
 
 var validateAddress = function (address) {
   if (address === null || address === undefined) {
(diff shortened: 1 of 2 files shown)


amousa11:
updated with tests

src/validate/validate-address.js
```diff
@@ -1,4 +1,21 @@
 "use strict";
+const createKeccakHash = require("keccak/js");
```
fanatid: why `keccak/js`?

test/validate/validate-address.js
```diff
@@ -0,0 +1,42 @@
...
+            assert.isTrue(validateAddress);
+        }
+    });
+    test({
+        description: "0xD1220A0cf47c7B9Be7A2E6BA89F429762e7b9aDb",
+        s: "0xD1220A0cf47c7B9Be7A2E6BA89F429762e7b9aDb",
+        assertions: function (validateAddress) {
+            assert.isTrue(validateAddress);
+        }
+    });
```
fanatid: Agree with @MicahZoltu there should be more tests, and they definitely should be moved to fixture...same code repeated a few times.

MicahZoltu:
Oh, I just realized you probably don't have merge rights @amousa11?  If not I can merge if I can trick CI into passing.

amousa11:
@MicahZoltu thank you! 

@fanatid i think i covered the issues you mentioned? Not sure what fixture is, however.

src/validate/validate-address.js
```diff
@@ -1,4 +1,21 @@
 "use strict";
+const sha3 = require('../utils/sha3');
```
MicahZoltu: The linter is erroring on this because `const` is not ES5 compatible.  Can you change this to `var`?  That should get CI passing, at which point I can merge.

MicahZoltu:
Looks like that change allowed the linter to run successfully over the whole file, and now there are a bunch more linter errors that it wants fixed:  https://travis-ci.org/ethereumjs/ethrpc/jobs/320015781#L2006-L2014

If you have the repo locally, I believe you can run the linter locally with
```
npm install
npm run lint
```

Alternatively, update the PR and CI will do it for you.  😄

MicahZoltu:
Thanks @amousa11!

amousa11:
Now #32 is taken care of as well! :)

PR #46, opened 2018-02-07
Make connect async

pgebheim:
Make `startBlockStream` actually act asynchronously when a `startingBlockNumber` is passed.

### DIFF
--- a/src/block-management/create-block-and-log-streamer.js
+++ b/src/block-management/create-block-and-log-streamer.js
@@ -41,7 +41,7 @@ var eth_getBlockByNumber = require("../wrappers/eth").getBlockByNumber;
  * @param {Configuration} configuration
  * @param {Transport} transport
  */
-function createBlockAndLogStreamer(configuration, transport) {
+function createBlockAndLogStreamer(configuration, transport, callback) {
   return function (dispatch, getState) {
     var blockNotifier = new BlockNotifier({
       getLatestBlock: transport.getLatestBlock,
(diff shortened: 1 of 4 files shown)


tinybike:
The changes look fine, but linter and tests are both failing.  Generally, you should run these prior to submitting a PR.

pgebheim:
Submitted the PR because a pile of tests fail for me that looked like they would only run under certain conditions so I wanted to check it out.

src/block-management/create-block-and-log-streamer.js
```diff
@@ -57,17 +56,30 @@ function createBlockAndLogStreamer(configuration, transport) {
     });
 
     internalState.setState({ blockAndLogStreamer: blockAndLogStreamer, blockNotifier: blockNotifier });
+
+    callback = callback ? callback : function (e) { if (e) console.log(e); };
```
tinybike: Suggest checking that the callback is actually a function (rather than just checking for existence):
```javascript
callback = typeof callback === "function" ? callback : function (e) { if (e) console.log(e); };
```

src/block-management/create-block-and-log-streamer.js
```diff
@@ -57,17 +56,30 @@ function createBlockAndLogStreamer(configuration, transport) {
...
     }
 
     if (typeof configuration.startingBlockNumber !== "undefined") {
-      var block = dispatch(eth_getBlockByNumber([configuration.startingBlockNumber, false]));
-      blockAndLogStreamer.reconcileNewBlockCallbackStyle(block, function (err) {
-        if (err) return console.error(err);
-        subscribeToBlockNotifier();
+      transport.getBlockByNumber(configuration.startingBlockNumber, function (err, block) {
+        if (err) {
+          return console.log(err);
```
tinybike: Shouldn't the callback be invoked here? i.e., `return callback(err)` instead of `return console.log(err)`

src/block-management/create-block-and-log-streamer.js
```diff
@@ -57,17 +56,30 @@ function createBlockAndLogStreamer(configuration, transport) {
...
+        }
+
+        blockAndLogStreamer.reconcileNewBlockCallbackStyle(block, function (err) {
+          if (err) {
+            if (callback) return callback(err);
+            return console.log(err);
```
tinybike: The callback is always defined now, so these two lines can be simplified to just `return callback(err)`.

src/block-management/ethrpc-transport-adapter.js
```diff
@@ -18,6 +18,9 @@ function createTransportAdapter(transporter) {
       getBlockByHash: function (hash, callback) {
         dispatch(eth.getBlockByHash([hash, false], errorSplittingWrapper(callback)));
       },
+      getBlockByNumber: function (n, callback) {
```
tinybike: Suggest using a more descriptive parameter name like `blockNumber`, instead of `n`.

src/block-management/create-block-and-log-streamer.js
pgebheim: Can is isFunction, and even get the out of band handler out of state...

src/block-management/ethrpc-transport-adapter.js
pgebheim: Good call

src/block-management/create-block-and-log-streamer.js
pgebheim: I didn't get add this whole file... re-pushing.

PR #50, opened 2018-02-23
2576 blockstream promises

tinybike:


src/block-management/create-block-and-log-streamer.js
```diff
@@ -52,32 +61,33 @@ function createBlockAndLogStreamer(configuration, transport, callback) {
...
-    internalState.setState({ blockAndLogStreamer: blockAndLogStreamer, blockNotifier: blockNotifier });
-
-    callback = isFunction(callback) ? callback : function (e) { if (e) console.log(e); };
-    function subscribeToBlockNotifier() {
-      blockNotifier.subscribe(function (block) {
-        blockAndLogStreamer.reconcileNewBlockCallbackStyle(block, function (err) { if (err) return console.error(err); });
+    var blockAndLogStreamer = new BlockAndLogStreamer(function (hash) {
+      return new Promise(function (resolve, reject) {
+        transport.getBlockByHash(hash, function (err, block) {
+          if (err) throw err;
```
MicahZoltu: I believe this should be:
```typescript
if (err) reject(err);
else resolve(block);`
```

MicahZoltu: If you aren't using async/await, you generally don't want to ever `throw`, because JS is terrible.

src/block-management/create-block-and-log-streamer.js
```diff
@@ -52,32 +61,33 @@ function createBlockAndLogStreamer(configuration, transport, callback) {
...
+        });
       });
-
-      callback(null);
-    }
-
+    }, function (filterOptions) {
+      return new Promise(function (resolve, reject) {
+        transport.getLogs(filterOptions, function (err, logs) {
+          if (err) throw err;
```
MicahZoltu: Same as above, don't throw, `reject(err)`.  Similarly, a couple lines down it should be `reject(new Error(...))`.

src/block-management/create-block-and-log-streamer.js
tinybike: Good catch (no pun intended)

PR #57, opened 2018-03-16
only allow a single eth connection

epheph:
This was the quickest way to do it. If this works correct, I'd like to revisit and refactor a bit, quite a bit of code exists to allow for multiple connections and deal with that, but we could likely remove quite a bit of code if this is the way we want to go


epheph:
fixing tests

MicahZoltu:
If there is no longer a requirement to simultaneously have multiple connections active with different capabilities (SYNC/non-SYNC), then I think this code could be cleaned/simplified a lot.  Recommend getting that question answered before deciding on a path forward.

epheph:
Yes, exactly, I wanted to make sure removing that functionality didn't break anything first for anything that depends on it. Do you know how it was envisioned to use multiple connections simultaneously? I see that it only ever uses the first eligible transport (hence creating our timeout problem)

MicahZoltu:
If I remember correctly, when you broadcast a message you tell it the capabilities the message requires (such as SYNC or DUPLEX).  It will then choose the first available transport that matches those capabilities, which is why the system supports multiple connection types.

tinybike:
@MicahZoltu I removed the message capability requirements parameter on the last round of ethrpc updates (when sync transport was also removed).  In theory we could use those for things like eth_subscribe that only work over full-duplex connections.  I took the parameter out b/c (10 we weren't actually using it that way, and (2) IMO it's much simpler for the transport layer to not have to worry about that concept.

epheph:
Ready for review

tinybike:
@epheph have you tested this with MetaMask?

src/transport/helpers/store-transport.js
```diff
@@ -0,0 +1,23 @@
...
+    Object.keys(transporter.internalState.reconnectListeners).forEach(function (key) {
+      transporter.internalState.reconnectListeners[key]();
+    });
+  });
+  transport.addDisconnectListener(function () {
+    Object.keys(transporter.internalState.disconnectListeners).forEach(function (key) {
+      transporter.internalState.disconnectListeners[key]();
+    });
+  });
+  onCompleteCallback(null, transporter);
```
tinybike: Would it be simpler for `storeTransport` to just return the transporter, instead of using a callback, since it doesn't do any asynchronous operations?

epheph: Totally. I mostly kept the original structure, but that's much better

src/transport/transporter.js
```diff
@@ -46,52 +44,63 @@ function Transporter(configuration, messageHandler, debugLogging, callback) {
...
-    }.bind(this));
-  }.bind(this));
-  configuration.httpAddresses.forEach(function (httpAddress, index) {
-    var httpTransport = new HttpTransport(httpAddress, configuration.connectionTimeout, messageHandler, function (error) {
-      resultAggregator.httpTransports[index] = (error !== null) ? null : httpTransport;
-      // TODO: propagate the error up to the caller for reporting
-      checkIfComplete(this, resultAggregator, callback);
+  // initiate connections to all provided addresses, only callback one that connects
+  someSeries([
+    function (callback) {
```
tinybike: Recommend not shadowing the `callback` param passed to `Transporter`.  Maybe `nextTransport` would be a good name?

tinybike: Same comment for each transport type btw

epheph: 👍

src/transport/transporter.js
```diff
@@ -46,52 +44,63 @@ function Transporter(configuration, messageHandler, debugLogging, callback) {
...
+      someSeries(configuration.httpAddresses,
+        function (httpAddress, next) {
+          var httpTransport = new HttpTransport(httpAddress, configuration.connectionTimeout, messageHandler, function (error) {
+            if (error !== null) return next(null);
+            return next(httpTransport);
+          });
+        },
+        callback);
+    }],
+    function (callback, next) {
```
tinybike: The use of `callback` in this function is confusing.  `callback` is actually the item in the array that you're iterating over, which is itself passed a parameter that (in the array) is also referred to as `callback`.

Suggestion: use `function (tryTransportType, nextTransportType)` instead.

src/transport/transporter.js
```diff
@@ -46,52 +44,63 @@ function Transporter(configuration, messageHandler, debugLogging, callback) {
...
+      callback(next);
+    }, function (foundTransport) {
+      if (!foundTransport) {
+        return callback(new Error("Unable to connect to an Ethereum node via any transport. (Web3, HTTP, WS, IPC)."), null);
```
tinybike: It looks like this was just copied from the old code, but usually the convention is to leave out the second (result) parameter if there's an error.

src/transport/helpers/store-transport.js
```diff
@@ -0,0 +1,23 @@
...
+function storeTransport(transporter, transport, onCompleteCallback) {
+  // subscribe to disconnect/reconnect callbacks for transport
+  if (!transport) return onCompleteCallback(new Error("Unable to connect to an Ethereum node via any transport. (Web3, HTTP, WS, IPC)."), null);
+  transporter.internalState.transport = transport;
+  transport.addReconnectListener(function () {
+    Object.keys(transporter.internalState.reconnectListeners).forEach(function (key) {
+      transporter.internalState.reconnectListeners[key]();
+    });
+  });
+  transport.addDisconnectListener(function () {
```
tinybike: I'm nearly positive that the callback for the disconnect listener should (or at least, _can_, depending on transport) have an `event` parameter, which should be passed to the individual listeners:
```javascript
transporter.internalState.disconnectListeners[key](event);
```

epheph:
@tinybike is this what you mean about fowarding the event? How would i test this? https://github.com/ethereumjs/ethrpc/pull/57/commits/c79916f50d0a05911a0e597db469eb6736647974

tinybike:
Yes, that's what I mean re: forwarding the event.  It looks like to test this you'd need to trigger a retryable error -- for websockets, that appears to be (in ws-transport.js):
```javascript
if (error.code === "INVALID_STATE_ERR") error.retryable = true;
if (error.message === "cannot call send() while not connected") error.retryable = true;
```
or
```javascript
if (this.webSocketClient.readyState === 3) {
```
Any of these should cause `notifyDisconnectListeners` (in abstract-transport.js) to be called, which receives the error object and passes it to the actual disconnect callbacks.

Note, there's a separate code path than the `onclose` handler defined in ws-transport.js, which also calls the disconnect callbacks and passes in an event object.  This is specifically triggered when a websocket closes, whereas the `notifyDisconnectListeners` stuff is triggered by unsuccessfully attempting to send an RPC request through an existing websocket (or over HTTP/IPC).

So...long story short, get the websocket into some screwed-up state, but still be connected, and then try to make an RPC request through it.  I do not know exactly how you would accomplish this though!  Maybe @MicahZoltu would know?  Most of this transport-layer stuff was his code originally :grin:

epheph:
Alright, I'll see if i can setup something like that. I also sent about and disabled that one IPC test while in CI

tinybike:
FWIW, I suspect it's more effort than it's worth to set that up, but if you can do it without a bunch of work go for it!

src/transport/abstract-transport.js
```diff
@@ -203,7 +203,7 @@ function notifyReconnectListeners(abstractTransport) {
 }
 
 /**
- * Notify all disconnect listeners of a reconect
+ * Notify all disconnect listeners of a reconnect
```
tinybike: Should say "disconnect"

PR #60, opened 2018-03-23
handle bad urls

epheph:
Danz found this bug. When an error occurs that is not async, it is thrown and the whole thing breaks, even if valid transports could be found later.


JohnDanz:
This looks like it should cover what i need 👍

PR #65, opened 2018-05-02
Add trezor as an account type

meowingtwurtle:


tinybike:
Lovely!

tinybike:
@MicahZoltu from ethrpc's standpoint that is true.  However, the UI needs separate labels so it knows what to display for the user, and it seems sensible to have these labels be the same as the labels that ethrpc uses internally.

PR #66, opened 2018-05-03
Use PollingBlockNotifier with MetaMask

tinybike:
Fixes https://app.clubhouse.io/augur/story/9797/new-blocks-not-appearing-in-ui-when-using-metamask

### DIFF
--- a/src/block-management/block-notifier.js
+++ b/src/block-management/block-notifier.js
@@ -3,18 +3,21 @@
 var Notifier = require("./notifier.js");
 var PollingBlockNotifier = require("./polling-block-notifier.js");
 var SubscribingBlockNotifier = require("./subscribing-block-notifier.js");
+var isMetaMask = require("../utils/is-meta-mask");
 
 function BlockNotifier(transport, pollingIntervalMilliseconds) {
   var blockNotifier;
   Notifier.call(this);
-
-  blockNotifier = new SubscribingBlockNotifier(transport, function () {
-    blockNotifier.destroy();
+  if (isMetaMask()) { // MetaMask doesn't throw an error on eth_subscribe, but doesn't actually send block notifications, so we need to poll instead...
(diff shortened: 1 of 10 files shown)


src/block-management/block-notifier.js
```diff
@@ -3,18 +3,21 @@
...
-    blockNotifier.subscribe(this.notifySubscribers);
-  }.bind(this));
+  } else {
+    blockNotifier = new SubscribingBlockNotifier(transport, function () {
+      blockNotifier.destroy();
+      blockNotifier = new PollingBlockNotifier(transport, pollingIntervalMilliseconds);
+      blockNotifier.subscribe(this.notifySubscribers);
+    }.bind(this));
+  }
   blockNotifier.subscribe(this.notifySubscribers);
```
epheph: Will this double-subscribe if it goes via the non-metamask but failed subscribe path?

src/utils/is-meta-mask.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+function isMetaMask() {
```
epheph: Use the same function here? https://github.com/ethereumjs/ethrpc/blob/7f63d147bdec1e17bab7618d28ed22f464c5ee6c/src/transport/web3-transport.js#L15

src/block-management/block-notifier.js
tinybike: Yep, but the first block notifier is destroyed, so it should be fine.  (It worked this way before, too.)

epheph: oh i see, yes, happens later  in the callback 👍

src/utils/is-meta-mask.js
tinybike: It's not quite the same function, but yes there's some duplicate code, added a global web3 check function.

PR #67, opened 2018-05-07
Tinybike/ch9853/new blocks are being processed twice

tinybike:
Resolves [ch9853](https://app.clubhouse.io/augur/story/9853/new-blocks-are-being-processed-twice).

I couldn't figure out what was causing the race condition described in 9853, due largely to the complexity of ethrpc's old block-management module.  I ended up just reimplementing the block-management module, and AFAICT the race condition(s) are no longer present.  @epheph please test-drive this PR and double-check that all's in good working order, since quite a few lines of code were changed here...


epheph:
Oh my. So I don't think I've experienced this in Augur node (or noticed? augur-node has a queue layer between block notification and processing). I wonder if it's specific to the transport? I want to make sure I'm testing the right path.

test/helpers.js
```diff
@@ -66,16 +66,8 @@ module.exports.getRpcConfiguration = function (transportType, transportAddress)
...
-        return "0x0000000000000000000000000000000000000b0b";
-      case "eth_gasPrice":
-        return "0x09184e72a000";
-      case "eth_subscribe":
-        return "0x00000000000000000000000000000001";
-      case "eth_unsubscribe":
-        return true;
-    }
+    if (request.method === "eth_gasPrice") return "0x09184e72a000";
+    // if (request.method === "eth_coinbase") return "0x0000000000000000000000000000000000000b0b";
```
MicahZoltu: Recommend deleting dead code.

package.json
```diff
@@ -6,17 +6,17 @@
...
     "link": "npm link speedomatic ethereumjs-blockstream",
     "genversion": "genversion src/version.js",
     "build": "mkdirp dist && browserify ./exports.js -o ./dist/ethrpc.js && uglifyjs ./dist/ethrpc.js -o ./dist/ethrpc.min.js",
-    "build:tests": "browserify ./test/reducers/*.js -o ./test/browser/bundle.js",
     "postinstall": "postinstall-build --only-as-dependency dist",
     "prepublishOnly": "postinstall-build dist",
     "preversion": "npm test",
-    "version": "npm run genversion && git add src/version.js"
+    "version": "npm run genversion && git add src/version.js",
+    "testwhy": "wtfnode ./node_modules/mocha/bin/_mocha --timeout 5000 ./test/stub-node-only.js"
```
MicahZoltu: This could use some documentation (a comment somewhere indicating what this does, why it exists, etc.)

package.json
```diff
@@ -6,17 +6,17 @@
   "license": "MIT",
   "main": "src/index.js",
   "scripts": {
-    "test": "mocha test/*.js test/block-management test/raw-transactions test/reducers test/transport test/utils",
-    "lint": "eslint src test/*.js test/block-management test/raw-transactions test/reducers test/transport test/utils",
+    "test": "mocha test",
+    "lint": "eslint src test",
```
MicahZoltu: I'm curious about this change.  Was it the old way because mocha wasn't recursive but the version bump of mocha adds support for recursive test locating?

epheph:
@tinybike I played around with how this behaves, it seems that we are still getting double blocks, but they are effectively deduping before trigger callback? But we don't necessarily know why we get double blocks?

package.json
tinybike: That's cruft left over from my testing actually, I'll delete it.

package.json
tinybike: No, both are recursive -- it's because we used to build a special browser testing bundle in the test directory, which wasn't intended to be run by mocha.  There's no reason to have that bundle just sitting in the directory though (should be generated as needed) so I removed it.

tinybike:
Hm, @epheph what do you mean by "effectively deduping before trigger callback"?  Where do you see double blocks coming in?

test/helpers.js
tinybike: Good catch.

epheph:
I took a quick look, nothing stood out, and i have a bit PR in augur-node i'd like to focus right now. 👍

PR #69, opened 2018-05-16
Fix memory leak

justinbarry:
https://app.clubhouse.io/augur/story/7926/augur-node-oom


package.json
```diff
@@ -35,10 +35,8 @@
     "ethereumjs-tx": "1.3.1",
     "immutable-delete": "1.0.1",
     "keccak": "1.2.0",
+    "lodash": "^4.17.10",
     "lodash.assign": "4.2.0",
```
MicahZoltu: Can you also remove lodash.assign?

package.json
```diff
@@ -35,10 +35,8 @@
...
-    "lodash.get": "4.4.2",
-    "lodash.set": "4.3.2",
-    "lodash.unset": "4.5.2",
```
MicahZoltu: What kind of impact will this change have on browser load times?

PR #80, opened 2018-07-06
Justinbarry/ch13255/only use web3transport if user specifically

justinbarry:
https://app.clubhouse.io/augur/story/13255/only-use-web3transport-if-user-specifically-requests-it

### DIFF
--- a/src/transport/transporter.js
+++ b/src/transport/transporter.js
@@ -104,7 +104,6 @@ function Transporter(configuration, messageHandler, debugLogging, callback) {
         nextTransport);
     },
     function (nextTransport) {
-      if (configuration.useWeb3Transport !== true) return nextTransport();
       var web3Transport = new Web3Transport(messageHandler, function (error) {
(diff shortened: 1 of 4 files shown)


src/transport/transporter.js
```diff
@@ -108,6 +103,13 @@ function Transporter(configuration, messageHandler, debugLogging, callback) {
         },
         nextTransport);
     },
+    function (nextTransport) {
+      if (configuration.useWeb3Transport !== true) return nextTransport();
```
tinybike: Web3Transport should be the lowest priority transport if useWeb3Transport isn't set to true, but it should still be used if nothing else is available.  I think this should only skip web3 transport altogether if useWeb3Transport is false AND there's already another transport selected.

tinybike:
Hm, this still isn't quite right, the UI is now defaulting to the backing node's coinbase account even after you select "Connect to MetaMask".  Not sure if the problem is here or in the UI changes though.

PR #84, opened 2018-07-10
remove resubmit logic

nuevoalex:
The resubmit logic here doesn't work in practice unless you're connected to a single dedicated node since the tx hash you get back likely wont exist elsewhere. This makes transactions continuously re-fire.

### DIFF
--- a/test/stub-node-only.js
+++ b/test/stub-node-only.js
@@ -1184,18 +1184,20 @@ describe("tests that only work against stub server", function () {
           });
         });
 
-        it("transact pool not accepting", function (done) {
-          function onSent() { }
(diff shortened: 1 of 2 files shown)


sharkcrayon:
Please close this CH story when this PR is accepted: https://app.clubhouse.io/augur/story/13529/txs-are-resubmitted-everywhere

test/stub-node-only.js
```diff
@@ -1184,18 +1184,20 @@ describe("tests that only work against stub server", function () {
           });
         });
 
-        it("transact pool not accepting", function (done) {
```
epheph: You wanted to keep this around?

PR #95, opened 2018-07-19
Added origin 'localhost' to websocket when using Nodejs

tinybike:


tinybike:
Please sanity check this @epheph

MicahZoltu:
This is a bug in geth.  The proposed solution is a workaround, but I recommend adding a comment in the code to remove it once Geth fixes their bug.  Here is the go-ethereum issue you can put in the comment:
https://github.com/ethereum/wiki/wiki/Block-Protocol-2.0/_compare/4f945460190997d62df9d56dd37b8084f039b75b...d4b0d720b79d5243956a713ef7a75f6b0d5d04ac

src/transport/ws-transport.js
```diff
@@ -22,7 +23,8 @@ WsTransport.prototype.connect = function (initialCallback) {
     initialCallbackCalled = true;
     initialCallback(err);
   };
-  this.webSocketClient = new WebSocketClient(this.address, [], undefined, undefined, { timeout: this.timeout }, this.websocketClientConfig);
+  var origin = isNode ? "localhost" : undefined;
```
MicahZoltu: I don't think you need the isNode check.  If this code is running in a browser any `Origin` header you set will get overwritten by the browser on the way out.

That being said, the `isNode` check does make it a tad more clear.  However, with a comment linking to the Geth bug I think the `isNode` check is overkill on adding clarity.

PR #98, opened 2018-07-26
Updated websocket origin header to exactly match browser

tinybike:


MicahZoltu:
From our conversation, it sounds like this is the right solution if the team sticks with port `8080`.  Per conversation though, recommend switching to a random port rather than `8080` to avoid conflict (also partially addresses a security bug presented on HackerOne).

tinybike:
@MicahZoltu tbh if we use a random port (instead of 8080) so that the port numbers don't match between Nodejs and browser, then the user has to whitelist both ports.  That isn't really any better than the situation now, where the user has to whitelist 127.0.0.1 and http://127.0.0.1:8080...

MicahZoltu:
I'm saying the browser should be using a random port so Augur App doesn't collide with other apps on the user's machine running web services.  `8080` is a defacto standard port, which means many services bind to that port (anything that needs to be accessed externally).  In this case, Augur App launches the link on behalf of the user, so they don't need to memorize it or type it, so there is no need (and in fact detriment) to using a defacto standard port.

I agree with you though, it sounds like for short term fix (until Geth fixes their bug) having node set `Origin` header to match whatever the browser sends is the correct option.

PR #105, opened 2018-09-28
Add disconnect function to ethrpc

adrake33:
Story details: https://app.clubhouse.io/augur/story/13939

### DIFF
--- a/src/disconnect.js
+++ b/src/disconnect.js
@@ -4,7 +4,7 @@ var destroyBlockandLogStreamer = require("./block-management/destroy-block-and-l
 var internalState = require("./internal-state");
 
 function disconnect() {
-  return function (dispatch, getState) {
+  return function (dispatch) {
     dispatch(destroyBlockandLogStreamer());
 
     internalState.get("transporter").close();
--- a/src/disconnect.js
+++ b/src/disconnect.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var destroyBlockandLogStreamer = require("./block-management/destroy-block-and-log-streamer");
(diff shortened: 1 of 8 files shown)


src/disconnect.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+var destroyBlockandLogStreamer = require("./block-management/destroy-block-and-log-streamer");
+var internalState = require("./internal-state");
+
+function disconnect() {
+  return function (dispatch) {
+    dispatch(destroyBlockandLogStreamer());
+
+    internalState.get("transporter").close();
```
epheph: Check that we have a transporter already ,so we don't call close on undefined?

epheph:
Seems like it should mostly work. The only question I have is about BlockStream. Specifically, this:

https://github.com/ethereumjs/ethrpc/blob/c0ca3fe36ffcaad50f30f6436560c31209457664/src/block-management/create-block-and-log-streamer.js#L10

Normally, new blocks are being fed to it, which causes it to take action (and call appropriate callbacks). 

However, when it is in backfill mode, it has noted a gap in blocks and fills in the middle, without ethrpc feeding in any blocks; it retrieves them itself. What happens when it is in this mode and a disconnect is called?

epheph:
Because the ethrpc is torn down, the blockstream no longer functions (because it isn't being fed logs) and any attempt by it to backfill stops (since the functions it received for getBlock and getLogs use an object that is no longer connected). In the future, it might be nice if blockstream had a sort of disconnect/reset as well, just something that wiped out filters and history.

PR #110, opened 2018-10-25
Downgrade stub server for test consistency

pgebheim:


MicahZoltu:
I'm curious what the problem with the newer version of `ethereumjs-stub-rpc-server` was?

PR #113, opened 2018-10-29
just warn on this since its fairly expected

nuevoalex:


src/block-management/start-polling-for-blocks.js
```diff
@@ -15,7 +15,7 @@ function startPollingForBlocks() {
       if (getState().debug.broadcast) console.log("Polling for latest block...");
       dispatch(eth_getBlockByNumber(["latest", false], function (err, block) {
         if (err) return console.error(err);
-        if (block === null) return console.error(new RPCError("BLOCK_NOT_FOUND"));
+        if (block === null) return console.warn(new RPCError("BLOCK_NOT_FOUND"));
```
MicahZoltu: `eth_getBlockByNumber('latest')` is "expected" to return `null`?  Under what conditions?

nuevoalex: `eth_getBlockByNumber` will return null whenever no block is found. Obviously `latest` _should_ make that case not occur but that doesn't seem to be the case.

There are some various issues suggesting this is very common in light nodes but we've observed it when hitting full nodes as well. It isn't an actual failure case for us as we just keep polling.

pgebheim: Unfortunately, I agree with Alex -- it's kinda like an expected unexpected behavior.

It is indeed on that telling the user about doesn't help since there's nothing they can do to it other than let us already recover.

so, I say YES to this patch :-D

PR #115, opened 2018-11-07
move to latest blockstream

epheph:


MicahZoltu:
I don't see how this can possibly be all the changes necessary to upgrade to blockstream 7.x?  The shape of the external interface changed, so more changes should be necessary...

epheph:
This is only the ethrpc side. It takes the filter object and blindly passes it to the getLogs function, so it doesn't need any update, and the actual code that integrates with it grabs the blockstream out of ethrpc to deal with it, so ethrpc doesn't change anything there either.

Need to update tests, though

PR #118, opened 2018-11-14
Ensure no race on pendingCallbacks

epheph:
There's a race here with pendingCallbacks in blockstream? The unsubscribe works, but the pending callback could have been queued already. Only happens on node6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
 
 node_js:
   - "8"
-  - "6"
 
 addons:
   apt:


MicahZoltu:
There shouldn't be a race here.  If you have multiple subscriptions, then they will all be fired even if you unsubscribe subscriber B from the callback of subscriber A.  However, if you unsubscribe from subscriber A in the callback for subscriber A you should not receive any future notifications, even if one is already in the pipe.

The callbacks are called [here](https://github.com/ethereumjs/ethereumjs-blockstream/blob/master/source/block-and-log-streamer.ts#L78-L81).  Notice that there is no async activity here, and since NodeJS is single threaded it means that code will collect all of the callbacks, and then call each of them back in sequence, with no continuations along the way.

Then if you look over [here](https://github.com/ethereumjs/ethereumjs-blockstream/blob/master/source/block-and-log-streamer.ts#L136-L139) you will notice that when unsubscribe happens it happens synchronously.

So in your original code, you'll receive the callback and then (synchronously) remove the subscription.  If you already have queued up callback on a new block it would not yet have entered `onBlockAdded`, thus it won't have read from the member variable that holds subscriptions yet.  When it does get called, it'll read the member variable and not see your subscription thus not call it back.

PR #119, opened 2018-11-15
Bump version

epheph:


MicahZoltu:
Same comment here as the other PR I think?

epheph:
Yes, this was done in 6.0.0, pushed to npm (so i'd like to merge to not lie about it), but i'll fix it in master.

PR #120, opened 2018-11-15
change how test is fixed

epheph:
Re #119

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
 
 node_js:
   - "8"
-  - "6"
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


MicahZoltu:
Doesn't this put the Node 6 failure back in?  Per our conversation, my vote would be to drop node 6 support as it is nearing end of life anyway.

Also consider dropping a console.log of the block in the callback to see if it is getting called back twice for the same log.

epheph:
#118 goes with it, prolly should have made them same PR.

petong:
@MicahZoltu is this safe to merge?

MicahZoltu:
Based on comments (which I don't remember the context of) it appears that node6 support is dropped, so this is fine now?

PR #122, opened 2018-11-27
sometimes  returns nonce as decimal, ethrpc couldn't handle it. This …

bthaile:
…occurs when Metamask is enabled and user is connected via Ledger

https://app.clubhouse.io/augur/story/18127/ledger-issue-can-t-create-market-and-can-t-get-rep-from-faucet

### DIFF
--- a/src/raw-transactions/set-raw-transaction-nonce.js
+++ b/src/raw-transactions/set-raw-transaction-nonce.js
@@ -16,10 +16,11 @@ var RPCError = require("../errors/rpc-error");
 function setRawTransactionNonce(packaged, address, callback) {
   return function (dispatch, getState) {
     dispatch(eth_getTransactionCount([address, "pending"], function (err, transactionCount) {
-      if (getState().debug.tx) console.log("transaction count:", address, transactionCount, parseInt(transactionCount, transactionCount.toString().startsWith("0x") ? 16 : 10));
+      if (getState().debug.tx) console.log("transaction count:", address, transactionCount);
       if (err) return callback(err);
       if (transactionCount == null) return callback(new RPCError("NO_RESPONSE"));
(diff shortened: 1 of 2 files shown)


epheph:
So the alternative implementation is
```
type transactionCount === number ? transactionCount : parseInt(transactionCount, 16))
```

Which might be better, I just don't think i'd know what to do if i got back a string that was not 0x prefixed? i guess not convert it? 🤷‍♂️ 

So i did a bit of digging on/how this is. So there's kinda two APIs, one is the JSON-RPC interface:
https://github.com/ethereum/wiki/wiki/JSON-RPC and https://wiki.parity.io/JSONRPC-eth-module.html#eth_gettransactioncount

which both indicate returning the response as a 0x-prefixed string.

Then there's also the web3/javascript-api library documentation:
https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethgettransactioncount

which indicates you get back a number

So it's usually up to the ethereum rpc library to do that conversion for you. The problem, I believe, is that metamask is handling both `eth_getTransactionCount` and `web3.eth.getTransactionCount` the same (with a code path that specifically intercepts calls to `eth_getTransactionCount`)

I think this a subtle metamask bug, i'll submit an issu. It's easy to detect when it returns a number, but no json-rpc implementations behave the same way as metamask overrides.

bthaile:
Don't we want 
typeof transactionCount === "number" ? parseInt(transactionCount, 16)) : transactionCount

we want to pass `verifyRawTransactionNonce` a hex value?

epheph:
No, verifyRawTransactionNonce wants a number, the `.hex()` does more than just plop a 0x on the front, i was wrong earlier

epheph:
https://github.com/MetaMask/metamask-extension/issues/5845

PR #125, opened 2018-12-12
changed block not found console error to a console warning

### LINKED ISSUE #417: Improve top-level error handling in ethrpc
chwy was seeing the following error while running the Augur App on Rinkeby:

We should make this and other top-level errors into warnings. We also need a less fragmented strategy for error handling/logging that works well across desktop & mobile.

JohnDanz:
closes https://github.com/AugurProject/augur-app/issues/417

I couldn't find the other one you referenced about `“asked for block x got block y”`


nuevoalex (approved):
Looks good! Will try to find the other errors that are really warnings

PR #126, opened 2018-12-12
Handle pump queue on retry without disconnect

epheph:


pgebheim:
Well I'm dumb. That's good to go.

PR #133, opened 2019-04-27
just poll instead of sometimes trying subscribe

nuevoalex:


src/block-management/listen-for-new-blocks.js
```diff
@@ -1,21 +1,10 @@
 "use strict";
 
 var startPollingForBlocks = require("./start-polling-for-blocks");
-var subscribeToNewBlockNotifications = require("./subscribe-to-new-block-notifications");
-var unsubscribeFromNewBlockNotifications = require("./unsubscribe-from-new-block-notifications");
```
MicahZoltu: Can these two files be deleted as well?  Are they referenced anywhere else?

PR #134, opened 2020-01-02
Improve HTTP Transport

pgebheim:
1) Only do net_version on initial connect (not reconnects)
2) Do not attempt to issue eth_subscribe over HTTP Transport
3) Fix tests


src/transport/http-transport.js
```diff
@@ -50,11 +51,24 @@ HttpTransport.prototype.close = function () {
...
+    });
+  }
+
   request({
     url: this.address,
     method: "POST",
     json: rpcObject, // lies! this actually wants a JSO, not a JSON string
     timeout: this.timeout,
+    headers: {
+      'User-Agent': 'ethrpc/'+version,
```
pgebheim: Oh should change this so it works in browser too

